PR #15, opened 2021-04-19
Feature/dependency free

BenceVirtonomy:
@Xiangyu-Hu Here is the pull request from GitHub. Please check if everything is fine for you. Some information:

- The Simbody and clapack source codes are included
- oneTBB is included as a submodule, there is an option in the main CmakeLists file to disable it, if someone wants to use the binaries instead of building it from here. After cloning the submodule, the oneTBB_edit.py needs to be run to make minor changes in the source.
- All 3D tests pass except the heart electromechanics one
- There is an option for Boost as well for the 2D cases


Xiangyu-Hu:
I have merged them.

PR #16, opened 2021-05-13
Dev pull request

BenceVirtonomy:
- New test case for contact between two soft bodies with different particle resolution
- Overloaded SolidBody constructor to allow resolution input for each soft body
- Method to find smallest time step relevant for the system


BenceVirtonomy:
@Xiangyu-Hu my changes from this week in the source code

PR #33, opened 2021-06-26
change action from simbody debian package to source

chuvirtonomy:
change action from simbody debian package to source


chuvirtonomy:
It is really absurd for me that we need to build the same source again again and again every time we change SPHinxsys code. i am looking for some possible artifact storage. Of course, if we have debian package working, that's the best..... will solve this in the future.

chuvirtonomy:
@Xiangyu-Hu  @BenceVirtonomy  This PR is ready. Please review and approve if you agree

BenceVirtonomy (approved):

PR #46, opened 2021-07-30
Mrge with bitbucket

Xiangyu-Hu:
Dear all,  here is the pull request for updated features from bitbucket.
The main changes are naming and typos. 
Beside that, the particle adaptation is modified to allow arbitrary refinement ratio related to the system resolution for bodies. For this, in the soft-body-contact case, the moving body resolution is 1.5 time (before only even integer number is allowed) respect to the system reference.
The solid model will still use the original numerical damping since I find it is much fast and actually more stable.
In addition, I have increase the difficulty of the twisting column case to a rate (300 rad/s) never be done in the literature. 
Please review it.
Thanks, 
Xiangyu


SPHINXsys/src/for_3D_build/geometries/geometry.cpp
```diff
@@ -124,22 +129,23 @@ namespace SPH
...
 	{
 		bool exist = false;
 		bool inside = false;
-		
-		for (auto& each_shape : triangle_mesh_shapes_)
+
+		for (auto &each_shape : triangle_mesh_shapes_)
 		{
-			TriangleMeshShape* sp = each_shape.first;
+			TriangleMeshShape *sp = each_shape.first;
```
BenceVirtonomy, line 148: Can we start refactoring the pointers to shared and unique pointers next time we have some changes for these pointers?

SPHINXsys/src/shared/bodies/body_relation.h
```diff
@@ -171,8 +174,8 @@ namespace SPH
 	class BaseBodyRelationContact : public SPHBodyRelation
 	{
 	protected:
-		StdVec<MeshCellLinkedList*> target_mesh_cell_linked_lists_;
-		StdVec<SearchRangeMultiResolution*> get_search_ranges_;
+		StdVec<CellLinkedList*> target_cell_linked_lists_;
```
BenceVirtonomy, line 177: For example unique pointers here, and these kind of places

SPHINXsys/src/shared/geometries/level_set.h
```diff
@@ -72,104 +69,111 @@ namespace SPH
...
+		void updateNormalDirectionForAPackage(LevelSetDataPackage *inner_data_pkg, Real dt = 0.0);
 		void updateKernelIntegrals();
-		void updateKernelIntegralsForAPackage(LevelSetDataPackage* inner_data_pkg, Real dt = 0.0);
-		void stepReinitializationForAPackage(LevelSetDataPackage* inner_data_pkg, Real dt = 0.0);
-		void markNearInterfaceForAPackage(LevelSetDataPackage* core_data_pkg, Real dt = 0.0);
-		void redistanceInterfaceForAPackage(LevelSetDataPackage* core_data_pkg, Real dt = 0.0);
-		virtual void initializeDataInACell(Vecu cell_index, Real dt) override;
-		virtual void initializeAddressesInACell(Vecu cell_index, Real dt) override;
-		virtual void tagACellIsInnerPackage(Vecu cell_index, Real dt) override;
+		void updateKernelIntegralsForAPackage(LevelSetDataPackage *inner_data_pkg, Real dt = 0.0);
```
BenceVirtonomy, line 145: also here

SPHINXsys/src/shared/materials/inelastic_solid.h
```diff
@@ -38,15 +38,15 @@ namespace SPH {
 	* @class PlasticSolid
 	* @brief Abstract class for a generalized plastic solid
 	*/
-	class PlasticSolid : public LinearElasticSolid
+	class PlasticSolid : public NeoHookeanSolid
```
BenceVirtonomy, line 41: why from NeoHookean?

SPHINXsys/src/shared/meshes/base_mesh.cpp
```diff
@@ -5,39 +5,65 @@
...
+				clamp((int)floor(rltpos[n] / grid_spacing_), 0, int(number_of_grid_points_[n]) - 2);
 		}
 		return cell_index;
 	}
 	//=================================================================================================//
-	Vecd BaseMesh::GridPositionFromIndex(Vecu grid_index)
+	Vecd BaseMesh::CellPositionFromIndex(const Vecu &cell_index)
+	{
+		Vecd cell_position;
+		for (int n = 0; n < cell_position.size(); n++)
```
BenceVirtonomy, line 49: size_t?

BenceVirtonomy (approved):

SPHINXsys/src/for_3D_build/geometries/geometry.cpp
msdrezavand: This is a good idea as smart pointers handle the creation and destruction of the object (replaces malloc/free new/delete).
We have them throughout the GPU version. 

The only thing that I am concerned about is weather the compiling units involved with SimBody are compatible with these advanced C++ features or not? In particular, the std::unique_ptr which has been introduced in C++14. The std::shared_ptr should be fine, as they are already shipped with C++11.

chuvirtonomy:
We have already used unique pointer in some classes

Get Outlook for Android<https://aka.ms/AAb9ysg>

SPHINXsys/src/for_3D_build/geometries/geometry.cpp
Xiangyu-Hu: Yes. My suggestion is use smart pointer for those non-performance related pointers. However, I would not suggest to use it for computing intensive cases.

SPHINXsys/src/shared/geometries/level_set.h
Xiangyu-Hu: The data Package is computing intensive.

SPHINXsys/src/shared/meshes/base_mesh.cpp
Xiangyu-Hu: SimTK Vecd return size as integer.

SPHINXsys/src/shared/bodies/body_relation.h
Xiangyu-Hu: For geometric modeling using smart pointer is a good idea.

SPHINXsys/src/shared/materials/inelastic_solid.h
Xiangyu-Hu: I think that plastic material derived from Neohookean better. Because plastic usually mean large deformation and large deformation, practically, can not be linear.

BenceVirtonomy: okey, understand

cases_test/test_3d_roof/src/3d_roof.cpp
```diff
@@ -149,7 +149,7 @@ int main()
 	TimeDependentExternalForce external_force(Vec3d(0.0, 0.0, gravitational_acceleration));
 
 	/** Creat a Cylinder body. */
-	Cylinder *cylinder_body = new Cylinder(system, "CylinderBody", new ParticleAdaptation(1.15, 0), new ParticleGeneratorDirect());
+	Cylinder *cylinder_body = new Cylinder(system, "CylinderBody", new ParticleAdaptation(1.15, 1.0), new ParticleGeneratorDirect());
```
msdrezavand, line 152: Also in such situations smart pointers come in handy, I believe.

PR #92, opened 2022-02-25
add boundary condition & examples

alundilong:
add examples:
#2D T shape in arbitrary direction
#3D T shape
#3D pipe with narrowness


Xiangyu-Hu:
The test is failed due to long run time?

SchrodingersAlphaGo:
fix

Xiangyu-Hu:
I think that the code should be able to pass ctest on the local machine before CI in Github.

Xiangyu-Hu:
Is the failed case can pass in your local machine?

SchrodingersAlphaGo:
> Is the failed case can pass in your local machine?

Yes, the failed case passed ctest on the local mechine.

Xiangyu-Hu:
OK. So you can switch off the regression case in 3d_dambreak case. then we have another try.

Xiangyu-Hu (approved):

PR #133, opened 2022-07-19
bug fix

ChiZhangatTUM:
The bug inc make will results compile error if the path contains "0".
The eigen requires c++17.


SPHINXsys/cmake/Dirsearch_shared.cmake
```diff
@@ -31,7 +31,7 @@ MACRO(SOURCE_DIRECTORIES_SHARED return_list)
         if(BUILD_WITH_SIMBODY)
             SET(dir_list ${dir_list} ${dir_path})
         else() # if Simbody is not built, ignore the simbody and clapack folders
-            if(NOT dir_list MATCHES ("/simbody/+" OR "/clapack_for_SPHinXsys/+") )
+            if(NOT (dir_list MATCHES "/simbody/+" OR  dir_list MATCHES "/clapack_for_SPHinXsys/+") )
```
BenceVirtonomy, line 34: You should remove this clapack_for_SPHinXsys, it doesn't exist anymore. We don't build clapack at all, it's handled in simbody. The necessary lapack function is replaced by eigen.

BenceVirtonomy (changes requested):

SPHINXsys/cmake/Dirsearch_shared.cmake
ChiZhangatTUM: Done, thanks.

BenceVirtonomy (approved):

CMakeLists.txt
```diff
@@ -83,7 +83,7 @@ elseif(EMSCRIPTEN)
     # set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${EM_FLAGS}")
 else()
     add_compile_definitions(__linux__=1)
-    set(CMAKE_CXX_FLAGS "-w -std=c++11 -DUNIX -pthread")
+    set(CMAKE_CXX_FLAGS "-w -std=c++17 -DUNIX -pthread")
```
FabienPean-Virtonomy, line 86: > [Eigen has good compiler support](https://eigen.tuxfamily.org/index.php?title=Main_Page#Compiler_support) as we run our test suite against many compilers to guarantee reliability and work around any compiler bugs. Eigen up to version 3.4 is standard C++03 and maintains reasonable compilation times. Versions following 3.4 will be C++14.

Eigen requires C++14 for now strictly speaking. But C++17 would be my preferred standard to target, I don't know if it will create side effects on our side though, but if CI passes here I'm fine with it

ChiZhangatTUM: Thanks for the response, please have a double check. 
Best
Chi

ChiZhangatTUM:
Check passed.

ChiZhangatTUM:
Wong action.

PR #135, opened 2022-07-22
Test cantilever beam with shell

FabienPean-Virtonomy:
## Description 
description

The cantilever beam can be found [here](https://en.wikipedia.org/wiki/Deflection_%28engineering%29#Uniformly-loaded_cantilever_beams). In the Euler-Bernoulli formalism, a rigid cross-section orthogonal to the centerline is assumed, so for isotropic linear elastic material, it is closest to the analytical value when Poisson's ratio is 0. In this test case, the maximum expected deflection $\delta=\tfrac{qL^4}{8EI}=7.246$ mm.

## Material properties: 
* Linear elastic 
  * E=21GPa (10x smaller than steel because too slow otherwise) 
  * nu=0.49, because it starts to fail somewhere between 0.48 and 0.49
* Density: 7800 (like steel, tested with 780 but does not changing anything)

## Results
|                       | Analytical | Abaqus (steel) | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys | SPHinxsys |
| --------------------- | ---------- | -------------- | --------- | --------- | --------- | --------- | --------- | --------- | --------- | --------- | --------- | --------- |
| Deflection (mm)       | 7.25       |                | 9.51      | 8.61      | 8.65      | 8.18      | 4.36      | 7.03      | 3.63      | 6.68      | 4.42      | 7.47      |
| Error (%)             | 0          | 1.17           | 31.3      | 18.8      | 19.3      | 12.9      | 39.9      | 2.99      | 50        | 7.78      | 39.0      | 3.09      |
| Runtime (s)           |            |                | 60        | 420       | 93        | 503       |           |           |           |           |           |           |
| Resolution (m)            |            |                | 0.004     | 0.002     | 0.004     | 0.002     | 0.004     | 0.002     | 0.004     | 0.002     | 0.004     | 0.002     |
| Cell or Node location |            |                | Cell      | Cell      | Node      | Node      | Cell      | Cell      | Node      | Node      | Node      | Node      |
| Extending BC          |            |                | FALSE     | FALSE     | FALSE     | FALSE     | TRUE      | TRUE      | TRUE      | TRUE      | TRUE      | TRUE      |
| Exact volume          |            |                | N.A.      | N.A.      | FALSE     | FALSE     | N.A.      | N.A.      | FALSE     | FALSE     | TRUE      | TRUE      |

## Discussion
It seems to converge close to analytical value (~3% error at dp=2mm), but it must be confirmed with simulation running at smaller resolution.
A main source of error is the boundary condition that must be carefully adjusted in order to have physical location x=0 to be fixed.
While writing this issue,  it also occurred to me that there is some shear of the cross-section as the shell formulation follows the Mindlin–Reissner theory, if I understood it correctly. So unless I constrain the pseudo-normal, the closest beam formulation to follow would actually be the Timoshenko beam theory. All things considered it might get close enough.

The pain point is when I run at lower Poisson ratio. Simulation fails already somewhere in-between 0.48 and 0.49. Your help here @DongWuTUM would be appreciated.


tests/3d_examples/cantilever_beam_with_shell/cantilever_beam_with_shell.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+void reset_parameters()
+{
+    rho = 7800/(scaling*scaling*scaling);
+    Youngs_modulus = 207e9/(scaling*scaling);
+    poisson = 0.49;
+    physical_viscosity = 1./ 4.0 * std::sqrt(rho * Youngs_modulus)*0.002*scaling;
+    ratio_fixed = 0.1;// Percent of body fixed
+    extend_bc = false;// Extend fixed bc by kernell cutoff
+    linear_load = 100./scaling;// 10 N/m
```
DongWuTUM, line 49: linear_load should be 100.

PR #136, opened 2022-07-26
Bugfix macos

ChiZhangatTUM:
The Bug of MAC OS compile error is fixed.


SPHINXsys/src/shared/particles/particle_sorting.h
```diff
@@ -33,6 +33,8 @@
 #include "base_data_package.h"
 #include "sph_data_containers.h"
 
+#define TBB_2021_2_0;
```
BenceVirtonomy, line 36: why do you need this? Please don't just add this flag like that. This flag caused issue before. Can you switch to the later version of TBB? Or use a Cmake option to define the flag if the user enables the option.

BenceVirtonomy (changes requested):
Hi Chi, please check the comment regarding the TBB flad. Otherwise it's fine for me.

BenceVirtonomy (approved):

PR #141, opened 2022-08-12
Fix simbody artefacts for CI

FabienPean-Virtonomy:
This is an attempt to fix the CI problem. Artefacts have been released on `Virtonomy/simbody` as a release which will remain valid permanently.


FabienPean-Virtonomy:
@Xiangyu-Hu Unlike what we discussed during the meeting, here is a quick solution that will avoid problems for quite a while

Xiangyu-Hu (approved):

PR #143, opened 2022-08-12
compile simbody onfly

Xiangyu-Hu:
compile simbody onfly to avoid read non open-source files


Xiangyu-Hu:
In case of future use.

PR #149, opened 2022-09-08
Xiangyu/to be merged to master

Xiangyu-Hu:
The major changes in the pull request: 
1. other than using std function, all particle dynamics local function applied in particle dynamics by using lambda function from, which is allowed to be inlined by compiler. 
3. different type of particle dynamics are explicitly specified in application by using template classes.
4. virtual functions in local function are mainly not used.
5. particle sorting taking out from fluid body and explicitly indicated.
6. some bug fixes.
7. Readme.md includes gif animations and links to source code.


DongWuTUM (approved):

ChiZhangatTUM (approved):

FabienPean-Virtonomy:
Are those changes in the API here to stay or only temporary ?
```cpp
SharedPtr<Gravity> gravity_ptr = makeShared<Gravity>(Vecd(0.0, -gravity_g));
SimpleDynamics<TimeStepInitialization> fluid_step_initialization(water_block, gravity_ptr);
ReduceDynamics<fluid_dynamics::AdvectionTimeStepSize> fluid_advection_time_step(water_block, U_max);
ReduceDynamics<fluid_dynamics::AcousticTimeStepSize> fluid_acoustic_time_step(water_block);
```

Xiangyu-Hu:
> 

Yes. I hope the type of particle dynamics to be explicit when they are applied. However, the naming of the them may be adjust for easy understanding.

FabienPean-Virtonomy:
This sounds like an implementation details for many which now causes extra overhead. What are the expected benefits from these changes ?

Xiangyu-Hu:
> This sounds like an implementation details for many which now causes extra overhead. What are the expected benefits from these changes ?

As there are only several type of the dynamics, I hope the overload is not much. 
The benefit is that people may write code with less errors and because you need to be aware which type of dynamics they are applying.

FabienPean-Virtonomy:
> As there are only several type of the dynamics, I hope the overload is not much. The benefit is that people may write code with less errors and because you need to be aware which type of dynamics they are applying.

Could you please give some example of errors ? The new way seems much more error-prone to me. It shifts the difficulty from the library developer to the library user because now one must now what type of class to use with a given class of interest. It looks quite difficult to know whether to use SimpleDynamics/ReduceDynamics/Dynamics1Level/ReduceAverage/ReduceInteraction/etc with a given concrete class (AcousticTimestep/StressRelaxation). 


> Another thing is that, because the particle dynamics using local dynamics as template parameter, without virtual function we seems not able to define a class combine the local dynamics and parallel looping in a single class.

One might get away with this problem using [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), but hard to say due to the overwhelming scope of the changes.

Xiangyu-Hu:
> > As there are only several type of the dynamics, I hope the overload is not much. The benefit is that people may write code with less errors and because you need to be aware which type of dynamics they are applying.
> 
> Is there some examples of errors previously remarked you could details ? The new way seems much more error-prone to me. It shifts the difficulty from the library developer to the library user because now one must now what type of class to use with a given class of interest. It looks quite difficult to know whether to use SimpleDynamics/ReduceDynamics/Dynamics1Level/ReduceAverage/ReduceInteraction/etc with a given concrete class (AcousticTimestep/StressRelaxation).
> 
> > Another thing is that, because the particle dynamics using local dynamics as template parameter, without virtual function we seems not able to define a class combine the local dynamics and parallel looping in a single class.
> 
> One might get away with this problem using [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), but hard to say due to the overwhelming scope of the changes.



> > As there are only several type of the dynamics, I hope the overload is not much. The benefit is that people may write code with less errors and because you need to be aware which type of dynamics they are applying.
> 
> Is there some examples of errors previously remarked you could details ? The new way seems much more error-prone to me. It shifts the difficulty from the library developer to the library user because now one must now what type of class to use with a given class of interest. It looks quite difficult to know whether to use SimpleDynamics/ReduceDynamics/Dynamics1Level/ReduceAverage/ReduceInteraction/etc with a given concrete class (AcousticTimestep/StressRelaxation).
> 
> > Another thing is that, because the particle dynamics using local dynamics as template parameter, without virtual function we seems not able to define a class combine the local dynamics and parallel looping in a single class.
> 
> One might get away with this problem using [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), but hard to say due to the overwhelming scope of the changes.

I am aiming at that, if choosing the particle dynamics wrongly, 
the code will not compile, because the mismatch of local dynamics to the type of particle dynamics. 
Now, we have obtained flexibility, on body part dynamics by using a template parameter but without modify the local dynamics.

SchrodingersAlphaGo (approved):

PR #158, opened 2022-11-03
Xiangyu/code refactory

Xiangyu-Hu:
In the this pull request, we have introduced the following updates:
1. the transport velocity formulation is not based on time step size anymore, and is generalized for multi-resolution simulation.
2. the level set shape now has sign correction to make sure correct signed distance probing.
3. a new dambreak case with particle  splitting and merging has been included.
4. a new flow around cylinder case with multi-resolution particle size has been included.
5. new class on life-time event has been introduced for further development.
6. geometric variable registration is done before particle generation. 
7. particle refinement now can be based on a shape which is not part the body shape.


FabienPean-Virtonomy:
Hi, I haven't found the part of the code referring to:
6. geometric variable registration is done before particle generation.
7. particle refinement now can be based on a shape which is not part the body shape.
Could you elaborate on those points please?

FabienPean-Virtonomy (approved):

Xiangyu-Hu:
> .

They are in the adaptation.h, line 83 and the class from 170.

PR #160, opened 2022-11-08
Add shell cases for test.

DongWuTUM:
### Observations & Following work
- The correction matrix B values for the spherical cap are strange; 
- The slightly nonuniform particle distribution results in unusual B values;
- **The calculation and application of B should be reconsidered;**
- For the pinched cylinder, where B is normal, the CFL number must be lower;
- **The numerical damping and hourglass control algorithm should be adjusted for the shell model.**
### Bug correction
- **A bug in the hourglass control algorithm is corrected**
### Test case 1: Pinched cylinder
**Reference: https://doi.org/10.1016/0168-874X(94)90097-3,  https://doi.org/10.1002/nme.2316**
- Initial particle positions are given manually;
- The bifurcation point of the resulting curve is accurately predicted;
- The details can be found here: https://www.researchgate.net/publication/361512276_An_efficient_SPH_formulation_for_plateshell_structures_with_finite_deformations#fullTextFileContent.

### Test case 2: Spherical cap
**Reference: Owen, D.R.J., 1980. Finite elements in plasticity, theory and practice,  https://doi.org/10.1007/s00466-017-1498-9**
- Ran with particle relaxation first;
- Calculation with B_ leads to the simulation crash;
- The result run with B_ removed is shown in the figure;
- The displacement pattern is similar.


DongWuTUM:
@BenceVirtonomy I create a new branch now that the master code has been updated and the bug in the shell particle reload has been fixed. (However, I later noticed that you had updated the branch feature/analytical 3d roof.)

BenceVirtonomy:
> @BenceVirtonomy I create a new branch now that the master code has been updated and the bug in the shell particle reload has been fixed. (However, I later noticed that you had updated the branch feature/analytical 3d roof.)

Great, thank you! The hourglass implementation was incorrect? I will pull this fix into my branch and test the hourglass again.
I'm testing with a more uniform particle distribution based on centroidal voronoi tessellation (CVT). Also, I ran the 3d roof case with the given lattice particles and refined particle distribution and it doesn't seem to converge either (at least not with a reasonable resolution, this is with the "particle_number" refined form 16 till 256). I think it's also due to the definition of the volume. I'm not sure if just using dp^2 can represent the whole geometry accurately.

image

BenceVirtonomy:
> @BenceVirtonomy I create a new branch now that the master code has been updated and the bug in the shell particle reload has been fixed. (However, I later noticed that you had updated the branch feature/analytical 3d roof.)

Is it possible that it's due to the damping? That is inversely related to the particle size, so I wonder if the damping is so high that the structure deforms less...

DongWuTUM:
Thank you, too! I calculated the roof case before. The results are shown in the figure. It seems to be converging. Anyway, I will test and double-check its convergence again, and check the definition of the volume. 
The physical damping is applied with a probability of 0.2 in each time step to prevent over-damping. I don't think it is the reason for the divergence.
image

BenceVirtonomy:
Did you check if it's already static with the 7e3 damping? I reduced the damping for myself because with 7e3 I get this curve:
image

BenceVirtonomy:
Okey, thanks. I would suggest plotting log x axis and keeping the refinement ratio constant.

Sent from Outlook for iOS<https://aka.ms/o0ukef>

DongWuTUM:
Thank you! Your suggestions and opinions are nice and useful.

DongWuTUM:
- I have recalculated the roof case. The results are shown in the Table. The first figure shows the displacement changing over time when particle_number = 256. 



- The convergence study shows the results are converging. The log scale for the x-axis should be just used for the displacement differences (or error if we know the analytical results) compared with the results when particle_number = 256. And we think the deformation and smooth length of the roof case are of the same order of magnitude, so it is not valuable to judge whether it converges. We also think the definition of volume is Okey and the volume is converging as the resolution rises. 


- We still need some time to check the calculation of B.

BenceVirtonomy:
That's great, thanks. I remember now I actually ran it with the B_ removed, that might be the difference for me. I will run it again and check.

DongWuTUM:
### Observations
- The algorithm is stable when the time step is in the order of 10e-7.

I examined the half-sphere case. When Bence tests it, it appears that the sphere is more than half.  So I give the gravity in another direction. 
Half_sphere

Xiangyu-Hu (commented):
All the new cases should have data for regression test.

Xiangyu-Hu:
> ### Observations
> * The algorithm is stable when the time step is in the order of 10e-7.
> 
> I examined the half-sphere case. When Bence tests it, it appears that the sphere is more than half. So I give the gravity in another direction. Half_sphere [ Half_sphere ](https://user-images.githubusercontent.com/100197018/207937298-f3e08260-29e7-4f45-9ba6-701ec8073d85.gif) [ ](https://user-images.githubusercontent.com/100197018/207937298-f3e08260-29e7-4f45-9ba6-701ec8073d85.gif)

This is beautiful !

BenceVirtonomy:
That's great, do you have a reliable time step calculation? @DongWuTUM

DongWuTUM:
> 

Not yet...

DongWuTUM:
> All the new cases should have data for regression test.

Okay.

BenceVirtonomy:
> > All the new cases should have data for regression test.
> 
> Okay.

Okay, but great that it's stable in general:)

BenceVirtonomy:
@DongWuTUM I agree that the sideways gravity is a better test. For me it's fine too but it gets hourglassed after a certain strain:
image
image

PR #165, opened 2022-11-15
Feature/CMake rework

FabienPean-Virtonomy:
Partial rework of the CMake files focusing on the libraries target sphinxsys_2d and sphinxsys_3d. C++ standard is more strictly enforced and CMake files simplified, so it should hopefully run on Windows without hiccups. The focus was on the build tree, not on the install tree.

Commits are explicit and atomic, but not necessarily independent (especially 4-7)

1. [filesystem is not experimental anymore in C++17](https://github.com/Xiangyu-Hu/SPHinXsys/commit/0cfef4c1ae5262007c5669b9f3f33afb3944b297)
2. [C++ standard conformance: always return something in non-void methods](https://github.com/Xiangyu-Hu/SPHinXsys/commit/a820715996806f874f710f9d3791575e249369d9)
3. [C++ standard conformance: pedantic](https://github.com/Xiangyu-Hu/SPHinXsys/commit/f8e3db5970e88558a1c108ba442aeabba4f5cf11)
4. [Rework CMake files for sphinxsys libraries](https://github.com/Xiangyu-Hu/SPHinXsys/commit/1e1e963721d88bfb7ff1884d02a1b634d80a573c)
5. [Cleanup now useless CMake files](https://github.com/Xiangyu-Hu/SPHinXsys/commit/b4947dd2de9a85184729af19d84b1721239b5c6e)
6. [Fix CMake target in examples](https://github.com/Xiangyu-Hu/SPHinXsys/commit/dc123895de435c3178d74fc3a809d692ccadfd52)
7. [Fix typo in TBB CMake target](https://github.com/Xiangyu-Hu/SPHinXsys/commit/ba467cb6d73f1bea72ddae8477fc326627fc45b4)
8. [Only add subdirectory if CMakeLists exists](https://github.com/Xiangyu-Hu/SPHinXsys/commit/564be425c3a5fb1e167bedf61e577bf1ee093323)


ChiZhangatTUM:
So, now we have to install Eigen?

FabienPean-Virtonomy:
Yes, for Linux it is easy as it is in every package manager, e.g. `apt install libeigen3-dev`

For Windows, you can install it with [vcpkg](https://github.com/microsoft/vcpkg)

ChiZhangatTUM:
> Yes, for Linux it is easy as it is in every package manager, e.g. `apt install libeigen3-dev`
> 
> For Windows, you can install it with [vcpkg](https://github.com/microsoft/vcpkg)

1, Why we abandon the header-only feature and use it as installed library?
2, NOT all users has the root authority to use "apt", it's nor always that easy.
3, When you do this kind of modification, at least give some instructions.

FabienPean-Virtonomy:
1. It is still header-only, just handled by a package manager. Eigen is ubiquitous, having local dependencies is in general a burden and better to have less of them. Though for Eigen, I agree it is not so critical due to its header-only nature.
2. Alternatively, vcpkg is cross-platform and in user-space. It will be phased in as the main source of dependency in the future.
3. My bad, I will add you as reviewer in the future as it would have been good to catch that during the review. I have been pointed to some documentation during the recurring meeting today.

This is only a partial revamp of the build system kept as limited as possible. The long-term goal is to lean it out. But I didn't want to make a bigger PR with the risk of being rejected as it requires a significant work to thin things down.

ChiZhangatTUM:
> 1. It is still header-only, just handled by a package manager. Eigen is ubiquitous, having local dependencies is in general a burden and better to have less of them. Though for Eigen, I agree it is not so critical due to its header-only nature.
> 
>     2. Alternatively, vcpkg is cross-platform and in user-space. It will be phased in as the main source of dependency in the future.
> 
>     3. My bad, I will add you as reviewer in the future as it would have been good to catch that during the review. I have been pointed to some documentation during the recurring meeting today.
> 
> 
> This is only a partial revamp of the build system kept as limited as possible. The long-term goal is to lean it out. But I didn't want to make a bigger PR with the risk of being rejected as it requires a significant work to thin things down.

My point is, we should make the installation process of SPHinXsys as simple as possible, implementing header-only library maybe one choice. Lots of users are not willing to update to the latest version is due to the complexity of installation.

FabienPean-Virtonomy:
Yes, that's obviously desirable and that's what I aim for. Lots of users, you mean this commit in particular or in general?

See #166 for fixes

ChiZhangatTUM:
> 

There are lots of master or phd student in china, they are using our code, while with windows, struggling with installation, and have no willingness to update to latest version.

PR #167, opened 2022-11-18
hotfix/cmake-boost

Bo-Zhang1995:
Change the boost find method in the CMake file.


BenceVirtonomy:
We are reworking the cmake to have a windows pipeline. Please wait with this PR until then.

PR #168, opened 2022-11-24
Correct the calculation of B for shell model

DongWuTUM:
1. Correct the calculation of B for the shell model;
2. Correct the bug in the hourglass control algorithm of the shell model;
3. Add 1 point Gaussian quadrature rule;
4. Simplify class "ConstrainShellBodyRegion".


SPHINXsys/src/shared/particle_dynamics/solid_dynamics/thin_structure_math.cpp
```diff
@@ -243,5 +243,30 @@ namespace SPH
...
+			return correction_matrix;
+		}
+		//=================================================================================================//
+		Mat3d getCorrectionMatrix(const Mat3d &local_deformation_part_one)
+		{
+			Mat2d reduced_local_deformation_part_one;
+			reduced_local_deformation_part_one[0][0] = local_deformation_part_one[0][0];
+			reduced_local_deformation_part_one[0][1] = local_deformation_part_one[0][1];
+			reduced_local_deformation_part_one[1][0] = local_deformation_part_one[1][0];
+			reduced_local_deformation_part_one[1][1] = local_deformation_part_one[1][1];
```
FabienPean-Virtonomy, line 259: +TinyReal on the diagonal ?

SPHINXsys/src/shared/particle_dynamics/solid_dynamics/thin_structure_math.cpp
```diff
@@ -243,5 +243,30 @@ namespace SPH
...
+			Mat2d reduced_correction_matrix; 
+			reduced_correction_matrix = SimTK::inverse(reduced_local_deformation_part_one); 
+			Mat3d correction_matrix(0.0);
```
FabienPean-Virtonomy, line 262: Is it fully safe to not have 1 on the diagonal at `correction_matrix[2][2]` ?

DongWuTUM: Because the shell model is a dimension reduction model, the correction_matrix[2][2] should be 0, not 1.

SPHINXsys/src/shared/particle_dynamics/solid_dynamics/thin_structure_math.cpp
DongWuTUM: From what I have tested, using 0 is fine.

FabienPean-Virtonomy: It would be preferable to have it uniform, so either removing it from 2D version or add it to the 3D

Xiangyu-Hu (approved):

BenceVirtonomy:
@DongWuTUM @Xiangyu-Hu I ran the stability test on my branch for the half sphere under self-weight, and I still get such out-of-plane displacements of the particles. Any idea why this happens?

image

image

Xiangyu-Hu:
> @DongWuTUM @Xiangyu-Hu I ran the stability test on my branch for the half sphere under self-weight, and I still get such out-of-plane displacements of the particles. Any idea why this happens?
> 
> image
> 
> image

Do you have the same issue if B is not used?

SPHINXsys/src/shared/particle_dynamics/solid_dynamics/thin_structure_math.cpp
DongWuTUM: Yes, you are right. I have removed TinyReal from the 2D model as we have initialized global_deformation with Eps.

DongWuTUM:
@BenceVirtonomy Hi, Bence. 

I can run the spherical cap case. Could you run the half-sphere case again without B, as Prof. Hu suggested?

And, as previously mentioned, you can try to reduce the time step because there are still stability issues from the numerical damping.

BenceVirtonomy:
It's much worse without the B_. You can try to run this half-sphere. It's fine until the the structure moves, this out-of-plane movement happens when the structure movement stops.

PR #170, opened 2022-11-25
Pin ubuntu version in CI

FabienPean-Virtonomy:
To avoid error rising in code because of newer TBB in Ubuntu 22.04 which is becoming the underlying version for Github-hosted runners


Xiangyu-Hu (approved):

FabienPean-Virtonomy:
The build step has passed successfully, it can be merged @Xiangyu-Hu

PR #172, opened 2022-11-28
Merge eigen with master

ChiZhangatTUM:
The main modifications is to implement the eigen, and other modifications. 
1,  Replace vector and matrix from SimTK with that from Eigen. 
2,  Split the function of SPHAdaptation::computeReferenceNumberDensity to 2D annd 3D folder to avoid  "ambiguous error".
3,  Move the OrthotropicSolid to 3D folder as it's generally 3D implementation. 
4,  Add new DFG funding NO. 
5,  Remove the eulerian test from ctest. 
6,  Remove author information form .cpp file, only kepp in .h file. Remove version information in .h files. 
7,  Replace ObserveBody with ProbeBody.


Xiangyu-Hu:
> The main modifications is to implement the eigen, and other modifications. 1, Replace vector and matrix from SimTK with that from Eigen. 2, Split the function of SPHAdaptation::computeReferenceNumberDensity to 2D annd 3D folder to avoid "ambiguous error". 3, Move the OrthotropicSolid to 3D folder as it's generally 3D implementation. 4, Add new DFG funding NO. 5, Remove the eulerian test from ctest. 6, Remove author information form .cpp file, only kepp in .h file. Remove version information in .h files. 7, Replace ObserveBody with ProbeBody.

Point 2, I do not think that there is ambiguous as the functions have different type of input parameters.
Ambiguous only happens when the function signatures are exact the same.

ChiZhangatTUM:
> 

When Matrix::Zero() is used as a default input parameter, ambiguity of overloaded function takes place due to the fact that Matrix::Zero() is not a matrix, but an Eigen expression.

FabienPean-Virtonomy:
> The main modifications is to implement the eigen, and other modifications. 1, Replace vector and matrix from SimTK with that from Eigen. 2, Split the function of SPHAdaptation::computeReferenceNumberDensity to 2D annd 3D folder to avoid "ambiguous error". 3, Move the OrthotropicSolid to 3D folder as it's generally 3D implementation. 4, Add new DFG funding NO. 5, Remove the eulerian test from ctest. 6, Remove author information form .cpp file, only kepp in .h file. Remove version information in .h files. 7, Replace ObserveBody with ProbeBody.

Thanks, it helps to have this list to guide the code review. Though it would be strongly preferable to not mix the changes and make a separate PR containing each one aspect (1-2-5/3/4/6/7).

ChiZhangatTUM:
> The alternative would be to evaluate the expression at the call site directly.
> 
> ```c++
> #include<Eigen/Eigen>
> using Vec2d = Eigen::Vector2<double>;
> using Vec3d = Eigen::Vector3<double>;
> using Vecd = Vec3d;
> void Foo(Vec2d v) {}
> void Foo(Vec3d v) {}
> int main() {
>     Foo(Vecd::Zero().eval());
>     return 0;
> }
> ```

Sure, I just choose the straightforward solution.

ChiZhangatTUM:
> 

Sorry for the inconvenience, I will follow your suggested principle next modification.

Xiangyu-Hu:
Another issue is: if you modify the observer class to probe, then the places relevant to the such change should be updated.
For example some places where the word 'probe' is already used should be changed to avoid misunderstanding. Basically, probe is for local value. Therefore, the global values obtaining by summation approaches should not be call probed value.

ChiZhangatTUM:
> Another issue is: if you modify the observer class to probe, then the places relevant to the such change should be updated. For example some places where the word 'probe' is already used should be changed to avoid misunderstanding. Basically, probe is for local value. Therefore, the global values obtaining by summation approaches should not be call probed value.

Ok, I will modify these in the following code refactoring.

SPHINXsys/src/for_3D_build/particles/solid_particles_supplementary.cpp
```diff
@@ -11,7 +11,7 @@ namespace SPH
 	{
 
 		Mat3d F = F_[particle_i];
-		Mat3d epsilon = 0.5 * (~F * F - Matd(1.0)); // calculation of the Green-Lagrange strain tensor
+		Mat3d epsilon = 0.5 * (F.transpose() * F - Matd::Identity()); // calculation of the Green-Lagrange strain tensor
```
BenceVirtonomy, line 14: @ChiZhangatTUM  there is a mix of Matd and Mat3d in this file

BenceVirtonomy (changes requested):

SPHINXsys/src/for_3D_build/particles/solid_particles_supplementary.cpp
ChiZhangatTUM: They are identical. 
There are mix usage of vector or matrix in dimensions, we can use a refactoring branch to clean them, and define a principle to follow.

SPHINXsys/src/for_3D_build/materials/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake) # main (top) cmake dir
+include(Headersearch)
+
+file(GLOB BASE_DIR_HEADERS *.h)
+DIR_INC_HEADER_NAMES("${BASE_DIR_HEADERS}" BASE_DIR_HEADER_NAMES)
+#message("${BASE_DIR_HEADER_NAMES}")
+INSTALL(FILES ${BASE_DIR_HEADER_NAMES} DESTINATION 3d_code/include)
```
FabienPean-Virtonomy, line 7: This file should be deleted, it is not needed anymore

tests/2d_examples/test_2d_elastic_gate/elastic_gate.cpp
```diff
@@ -158,7 +158,7 @@ int main()
 	gate.defineParticlesAndMaterial<ElasticSolidParticles, SaintVenantKirchhoffSolid>(rho0_s, Youngs_modulus, poisson);
 	gate.generateParticles<ParticleGeneratorLattice>();
 
-	ObserverBody gate_observer(system, "Observer");
+	ProbeBody gate_observer(system, "Observer");
 	gate_observer.defineAdaptationRatios(1.15, 2.0);
 	gate_observer.generateParticles<ObserverParticleGenerator>(observation_location);
```
FabienPean-Virtonomy, line 163: ProbeBody still requires ObserverParticles and Observer related stuff

FabienPean-Virtonomy (commented):
I think I am done going over it now. 

Changing the name of ObserverBody to ProbeBody is a breaking change with no functional benefit, without clear motivation, and partially implemented. I would recommend removing those changes from here and make a future PR more complete.

tests/2d_examples/test_2d_elastic_gate/elastic_gate.cpp
ChiZhangatTUM: I have reverse the ProbeBody to ObserverBody.

ChiZhangatTUM:
Now the ProbeBody and the adaptation has been reversed.

SPHINXsys/src/for_3D_build/materials/elastic_solid_supplementary.cpp
```diff
@@ -0,0 +1,88 @@
+#include "elastic_solid.h"
```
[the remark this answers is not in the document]
ChiZhangatTUM: According to its CalculateAllLambda() function, this is generally 3D. It may be fully removed to for_3D_build.

ChiZhangatTUM (commented):
We should be very careful when use un-initialized variables. 
Lots of bugs come from this issue.

Xiangyu-Hu:
I am happy to see active discussions on pull request.

FabienPean-Virtonomy:
> We should be very careful when use un-initialized variables. Lots of bugs come from this issue.

Indeed, I used it in that context because Vec2d/Vec3d is only used for tag dispatch and overload resolution, not in the body of the method. I'm not sure it is the best solution to the problem (your initial choice of splitting might be cleaner in the end). But at this point, better trying to wrap up the PR.

Note that the problem with Orthotropic material is not resolved and will come bite soon again depending on how the library is built (static vs shared). It means either offloading the whole implementation in `elastic_solid_supplementary.h` but then the material is not found by including `all_materials.h` without further changes, or building a dummy 2d version to preserve symmetry.

ChiZhangatTUM:
> 
I can reverse it back to the shared folder, while all vector and matd contained should be change to its 3D form. 
However, the logic is wrong in this case, as a function in shared folder is only works for 2D modeling.

ChiZhangatTUM:
Thanks for the effort.

FabienPean-Virtonomy:
@ChiZhangatTUM I made a fix following the existing convention. Dimension-specific functions are implemented in each version of the library. The 2D version is reporting an error since not fully implemented. The other functions have been made dimension- agnostic but required to `constexpr` some variables (shouldn't be a problem as they should be constexpr anyway). If this is ok with you, then on my side it is fine to merge it after CI passes.

FabienPean-Virtonomy (approved):

ChiZhangatTUM:
> @ChiZhangatTUM I made a fix following the existing convention. Dimension-specific functions are implemented in each version of the library. The 2D version is reporting an error since not fully implemented. The other functions have been made dimension- agnostic but required to `constexpr` some variables (shouldn't be a problem as they should be constexpr anyway). If this is ok with you, then on my side it is fine to merge it after CI passes.

Thanks for the work. 
The present form looks good. 
We can merge it after all the tests pass. 
Thanks for the cooperation.

ChiZhangatTUM:
@BenceVirtonomy  the quested change is not essential, would you please withdraw it, and then we can merge.

BenceVirtonomy (approved):

BenceVirtonomy:
Approved, thank you!

PR #174, opened 2022-12-02
Feature/buildsystem rework

FabienPean-Virtonomy:
The documentation took longer to write than expected. I could not test the procedure myself; please test them on your side before considering merging.

The tests passed on our fork, see [here](https://github.com/Virtonomy/SPHinXsys/actions/runs/3601091517). So, it should hopefully be passing here as well.

Several remarks:
* Sticking to Ubuntu 20.04 to avoid altering cpp files. This should be done in another step.
* Cpp files altered modified to respect C++17
  * Removing Boost.Filesystem because `std::filesystem` is part of C++17
  * [particle_generator_network.cpp](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...Virtonomy:SPHinXsys:feature/buildsystem-rework#diff-3ff20a60647ac82b80fa1251de2be67d907ac55d5a41b7ed9a7e531de5842828) to remove `std::random_shuffle` deleted in C++17
* Windows building shared libraries on CI
  * Hence moving `physical_time_` from source to header as static inline variable (C++17)
* Heavy cleaning of CMakeLists files
  * 2D examples are left to do (simply too many of them, I focused on 3D)
  * Dependencies are not based on submodules anymore 
  * Dependencies externally managed through [microsoft/vcpkg](https://www.github.com/microsoft/vcpkg)
  * Install targets not finalized. It is fully assumed that the development is done via the buildtree
  * Reduce the amount of custom cmake macro/function. Further cleanup to be done
* MacOS is left out
* Both `test_2d_eulerian...` disabled on Windows, `test_2d_eulerian_flow_around_cylinder` disabled for Linux


Xiangyu-Hu (approved):

ChiZhangatTUM:
Hi @FabienPean-Virtonomy ,
In my case, cmake can find gtest, while report the following error, any idea?

CMake Error at SPHINXsys/src/CMakeLists.txt:32 (add_library):
  Target "sphinxsys_3d" links to target "GTest::gtest" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?

FabienPean-Virtonomy:
> Hi @FabienPean-Virtonomy , In my case, cmake can find gtest, while report the following error, any idea?
> 
> CMake Error at SPHINXsys/src/CMakeLists.txt:32 (add_library): Target "sphinxsys_3d" links to target "GTest::gtest" but the target was not found. Perhaps a find_package() call is missing for an IMPORTED target, or an ALIAS target is missing?

I need more information: did you install gtest via vcpkg? Did you pass the CMAKE_TOOLCHAIN_FILE variable to cmake like shown in the installation guide?

PR #176, opened 2022-12-06
List out all SIMD flags on Linux systems instead of `-march=native`

FabienPean-Virtonomy:
Let's see if "Illegal" instructions during tests disappear


FabienPean-Virtonomy:
The CI passes for Linux, but since the problem does not occur every time, we can only see over time.

Xiangyu-Hu:
We can let the action job rerun several time to have a look.

Xiangyu-Hu:
> The CI passes for Linux, but since the problem does not occur every time, we can only see over time.

Then, which one is chosen for compiling?

FabienPean-Virtonomy:
> We can let the action job rerun several time to have a look.

At best, this solves the problem, at worst it does not change anything. I will run it one or two more times today to fulfill your request.

> Then, which one is chosen for compiling?

The block that is deleted in this PR was overriding a variable that contained individual SIMD feature flags (`-mavx2 -mfma -mavx`, etc) by a simpler `-march=native`.

Xiangyu-Hu:
> > We can let the action job rerun several time to have a look.
> 
> At best, this solves the problem, at worst it does not change anything. I will run it one or two more times today to fulfill your request.
> 
> > Then, which one is chosen for compiling?
> 
> The block that is deleted in this PR was overriding a variable that contained individual SIMD feature flags (`-mavx2 -mfma -mavx`, etc) by a simpler `-march=native`.

As a CPU may fit several flags, which is chosen finally then?

FabienPean-Virtonomy:
> As a CPU may fit several flags, which is chosen finally then?

The CMake file tests for SIMD features and adds flags depending on availability. The compiler uses all instructions set provided. I edited the title to reflect more accurately what is now happening.

Xiangyu-Hu (approved):

ChiZhangatTUM:
@FabienPean-Virtonomy 
Without VCPKG installed gtest, I have to remove the "CONFIG" in find_package(GTest CONFIG REQUIRED) in cmake, 
Its OK?

FabienPean-Virtonomy:
> @FabienPean-Virtonomy Without VCPKG installed gtest, I have to remove the "CONFIG" in find_package(GTest CONFIG REQUIRED) in cmake, Its OK?

The goal of vcpkg is to uniformize consuming dependencies across OSs and different versions, all of this in user-space. Please do not modify the CMakeLists and use vcpkg instead.

ChiZhangatTUM:
> 



> > @FabienPean-Virtonomy Without VCPKG installed gtest, I have to remove the "CONFIG" in find_package(GTest CONFIG REQUIRED) in cmake, Its OK?
> 
> The goal of vcpkg is to uniformize consuming dependencies across OSs and different versions, all of this in user-space. Please do not modify the CMakeLists and use vcpkg instead. If users face problems with vcpkg, then we can look further to find a solution and adapt then. Otherwise fitting the CMakeLists for all individual possibilities/needs will end up creating the monster that existed before

To be honest, I do not like the way of relying very thing on a specific package manager, resulting the total installation procedure to be a black-box. Anyway, if you all think that this is more convenient, I'm going to reserve my opinion.

PR #181, opened 2022-12-12
Add CI workflow for MAC OS. 

ChiZhangatTUM:
1, Workflow for MAC OS is added. 
2, Some virtual function add "override" for consistency.
3, Small number (x.e-3xx) of reference data for regression test in test_2d_diffusion is change to "0".
 Points 2 and 3 are essential for successful build and test pass.


ChiZhangatTUM:
Looks like we need gcc@9, 10 or 11 to get GNU Fortran. 
https://github.com/actions/runner-images/blob/main/images/macos/macos-10.15-Readme.md

gcc-9 (Homebrew GCC 9.5.0) 9.5.0 - available by gcc-9 alias
gcc-10 (Homebrew GCC 10.4.0) 10.4.0 - available by gcc-10 alias
gcc-11 (Homebrew GCC 11.3.0) 11.3.0 - available by gcc-11 alias
GNU Fortran (Homebrew GCC 9.5.0) 9.5.0 - available by gfortran-9 alias
GNU Fortran (Homebrew GCC 10.4.0) 10.4.0 - available by gfortran-10 alias
GNU Fortran (Homebrew GCC 11.3.0) 11.3.0 - available by gfortran-11 alias

ChiZhangatTUM:
brew reinstall gcc works for getting gfortran. 
Looks like the "Install dependency" part is very computational expensive, change brew to install simbody maybe an alternative.

Xiangyu-Hu:
> 1, Workflow for MAC OS is added. 2, Some virtual function add "override" for consistency. 3, Small number (x.e-3xx) of reference data for regression test in test_2d_diffusion is change to "0". Points 2 and 3 are essential for successful build and test pass.

What is the problem for the point 3?

ChiZhangatTUM:
> > 1, Workflow for MAC OS is added. 2, Some virtual function add "override" for consistency. 3, Small number (x.e-3xx) of reference data for regression test in test_2d_diffusion is change to "0". Points 2 and 3 are essential for successful build and test pass.
> 
> What is the problem for the point 3?

This kind of small number , like x.xe-323 , can’t be converted from string to double with clang compiler.

Xiangyu-Hu:
7777777

> > > 1, Workflow for MAC OS is added. 2, Some virtual function add "override" for consistency. 3, Small number (x.e-3xx) of reference data for regression test in test_2d_diffusion is change to "0". Points 2 and 3 are essential for successful build and test pass.
> > 
> > 
> > What is the problem for the point 3?
> 
> This kind of small number , like x.xe-323 , can’t be converted from string to double with clang compiler.

Rewriting the out file by hand seems not a good option.
The possible solution is format the accuracy of the output so that small number is written as zero?

ChiZhangatTUM:
> 

Sure , this can be combined with the bug fix of the regression test in another PR.

Xiangyu-Hu:
> > 
> 
> Sure , this can be combined with the bug fix of the regression test in another PR.

Better do not keep a tail for the PR. 
Actually, it should be done easily, as from the simbody documentation.
---
/** Format a double as a printable %String. Nonfinite values are formatted as
NaN, Inf, or -Inf as appropriate (Matlab compatible). The default format
specification includes enough digits so that the identical value will be
recovered if the string is converted back to double. **/
SimTK_SimTKCOMMON_EXPORT explicit String(double r, const char* fmt="%.17g");
---
just need to change the default format to a less accurate version.

FabienPean-Virtonomy:
> brew reinstall gcc works for getting gfortran. Looks like the "Install dependency" part is very computational expensive, change brew to install simbody maybe an alternative.

This step is cached after a successful build (hopefully it will be cached even on failure of some other step in a future version of the gihub action). The hydra of OS is finally complete, and potentially the one of compilers too. That's great!

ChiZhangatTUM:
> > brew reinstall gcc works for getting gfortran. Looks like the "Install dependency" part is very computational expensive, change brew to install simbody maybe an alternative.
> 
> This step is cached after a successful build (hopefully it will be cached even on failure of some other step in a future version of the gihub action). The hydra of OS is finally complete, and potentially the one of compilers too. That's great!
> 
> I saw that illegal instructions popped up again for the Linux CI. I made an issue to keep track of it, see #182 . Finding the source of problem might be a sinkhole of time.

That’s great.
Thanks for the optimization.

ChiZhangatTUM:
> 

Sure,  I will revise it tomorrow. 
BTW, some thing wrong with the test? All tests passed in my last draft PR.

FabienPean-Virtonomy:
> BTW, some thing wrong with the test? All tests passed in my last draft PR.

One example on windows hung up (that happened once in the past, but too rare to establish diagnostic) and macOS example `test_2d_eulerian_flow_around_cylinder` didn't converge to tolerance. This example is disabled for both Linux and Windows, not macOS yet. Restarting them should suffice.

ChiZhangatTUM:
> > BTW, some thing wrong with the test? All tests passed in my last draft PR.
> 
> One example on windows hung up (that happened once in the past, but too rare to establish diagnostic) and macOS example `test_2d_eulerian_flow_around_cylinder` didn't converge to tolerance. This example is disabled for both Linux and Windows, not macOS yet. Restarting them should suffice.

OK. Thanks.

ChiZhangatTUM:
> > brew reinstall gcc works for getting gfortran. Looks like the "Install dependency" part is very computational expensive, change brew to install simbody maybe an alternative.
> 
> This step is cached after a successful build (hopefully it will be cached even on failure of some other step in a future version of the gihub action). The hydra of OS is finally complete, and potentially the one of compilers too. That's great!
> 
> I saw that illegal instructions popped up again for the Linux CI. I made an issue to keep track of it, see #182 . Finding the source of problem might be a sinkhole of time.

We may try intel compiler in the future work.

ChiZhangatTUM:
Ready to be merged.

FabienPean-Virtonomy (approved):

PR #188, opened 2023-01-12
Set CI to use macos-12

FabienPean-Virtonomy:


FabienPean-Virtonomy:
The problem might actually have been that the name was not correct for setting the runner image (macOS-latest vs macos-latest) but it is still better to fix the version in general (or have a matrix that tests for several versions)

Xiangyu-Hu (approved):

PR #194, opened 2023-01-27
[BUG] Channel flow with free stream velocity correction

anastazja-bro:
FreeStreamVelocityCorrection is causing unexpected changes in the density distribution and incorrect velocity profile near the outlet


Xiangyu-Hu:
How long is the runtime of this test?

anastazja-bro:
The test is set for 200 sec. of physical time, but the flow disturbance is visible already after 1-2 sec. which took around 1 min to simulate on my PC

Xiangyu-Hu:
> The test is set for 200 sec. of physical time, but the flow disturbance is visible already after 1-2 sec. which took around 1 min to simulate on my PC

Does problem solved?

YuVirtonomy:
> > The test is set for 200 sec. of physical time, but the flow disturbance is visible already after 1-2 sec. which took around 1 min to simulate on my PC
> 
> Does problem solved?

This problem is solved.

- Need to add TimeDependentAcceleration
- Should not use parabolic flow for the inlet but uniform flow.
I think this pull/issue should be closed

Xiangyu-Hu:
> > > The test is set for 200 sec. of physical time, but the flow disturbance is visible already after 1-2 sec. which took around 1 min to simulate on my PC
> > 
> > 
> > Does problem solved?
> 
> This problem is solved.
> 
> * Need to add TimeDependentAcceleration
> * Should not use parabolic flow for the inlet but uniform flow.
>   I think this pull/issue should be closed

Could you make a ci test in the pull request ao that we can merge it to master. Thanks.

PR #198, opened 2023-02-03
shell contact algorithm

DongWuTUM:
Shell contact algorithm is updated.


BenceVirtonomy:
@DongWuTUM Can you please give more details in the PR description as to what changes - how that influences the usage?

DongWuTUM:
> @DongWuTUM Can you please give more details in the PR description as to what changes - how that influences the usage? Also images from the test case result?

Yes. The maxium contact density is 0.005 * density. The factor, sigma/contact_max_, means the ratio of the calculated contact density to the maximum density. The purpose of this updated algorithm is to prevent penetration and simplify the shell contact model. The usage and test case results are essentially unchanged.

DongWuTUM:
> contact_max_

DongWuTUM:
> `0.005*density` was there previously, but why actually this magic number, heuristics or theory based ?
> 
> On the PR itself, what is the gaussian integration discarded ?

The shell contact model was last updated with Eigen. However, it was not the final version. So I'm updating it now.

Looking at the class `ContactDensitySummation`, the maximum contact pressure is `density * c * c` (`c` is the sound wave speed), i.e., `K` (bulk modulus). 
The contact pressure `K` is extremely huge. Ihe shell model, a layer of particles and a huge contact pressure result in a single particle having a huge contact acceleration, which is one of the sources of instability. Thus, we add a factor `0.005` to maintain enough contact pressure to prevent penetration while also maintaining stability. As a result, the factor `0.005` is determined by theory and numerical experiments.

By using the Gauss-Legendre quadrature integration in the previous shell contact model, we want to add Gaussian points around each particle (2 points for 2D problems and 8 points for 3D problems, same case when calculating the `contact_max_`) to also ensure enough contact pressure and improve stability. However, numerical experiments show that we don't need to do this because the results are alomost the same with or without the additional Gaussian points.

FabienPean-Virtonomy:
> Thus, the factor 0.005 is determined by theory and numerical experiments.

According to your description, it is purely heuristics then. On how many cases was it tested (2D, 3D) ?

> numerical experiments show that we don't need to do this because the results are almost the same with or without the additional Gaussian points.

Good to hear, simpler is better ! Though again, on how many cases was it tested (2D, 3D) ?

SPHINXsys/src/shared/particle_dynamics/solid_dynamics/contact_dynamics.cpp
```diff
@@ -129,6 +126,26 @@ namespace SPH
...
+			for (size_t k = 0; k < contact_configuration_.size(); ++k)
+			{
+				StdLargeVec<Real>& contact_Vol_k = *(contact_Vol_[k]);
+				Neighborhood& contact_neighborhood = (*contact_configuration_[k])[index_i];
+				for (size_t n = 0; n != contact_neighborhood.current_size_; ++n)
+				{
+					Real corrected_W_ij = std::max(contact_neighborhood.W_ij_[n] - offset_W_ij_[k], 0.0);
+					sigma += corrected_W_ij * contact_Vol_k[contact_neighborhood.j_[n]];
+				}
+				contact_density_i += 0.005 * sigma * calibration_factor_[k];
```
FabienPean-Virtonomy, line 144:
```suggestion
				constexpr Real heuristic_limiter = 0.005;
				contact_density_i += heuristic_limiter * sigma * calibration_factor_[k];
```

DongWuTUM:
> > Thus, the factor 0.005 is determined by theory and numerical experiments.
> 
> According to your description, it is purely heuristics then. On how many cases was it tested (2D, 3D) ?
> 
> > numerical experiments show that we don't need to do this because the results are almost the same with or without the additional Gaussian points.
> 
> Good to hear, simpler is better ! Though again, on how many cases was it tested (2D, 3D) ?

Two cases are put to the test. Could you help us to test more?

BenceVirtonomy:
> Two cases are put to the test. Could you help us to test more?

We can sure, please describe the tests and the results, then we can look to add more.

DongWuTUM:
> > Two cases are put to the test. Could you help us to test more?
> 
> We can sure, please describe the tests and the results, then we can look to add more.

The two test cases, ball_shell_collosion and 3d_elasticSolid_shell_collision, are available. You can run them and have a check.

BenceVirtonomy:
> The two test cases, ball_shell_collosion and 3d_elasticSolid_shell_collision, are available. You can run them and have a check.

Do you also take the thickness into consideration now concerning the distance needed to reach contact between two bodies?
I think last time you explained that thickness = resolution is ideal for shell but it would make sense to allow for at least thickness = 0.5-2x resolution.

DongWuTUM:
> > The two test cases, ball_shell_collosion and 3d_elasticSolid_shell_collision, are available. You can run them and have a check.
> 
> Do you also take the thickness into consideration now concerning the distance needed to reach contact between two bodies? I think last time you explained that thickness = resolution is ideal for shell but it would make sense to allow for at least thickness = 0.5-2x resolution.

Not yet, although we have considered the resolution difference between two contact bodies. 
I agree that it makes sense to allow for at least thickness equal to 0.5-2.0 * resolution. 
The effect of estimated contact distance error is usually minimal as the thickness is small. 
It does, however, have an effect in some cases. We will factor thickness into the contact distance in the near future.

BenceVirtonomy:
> Not yet, although we have considered the resolution difference between two contact bodies. I agree that it makes sense to allow for at least thickness equal to 0.5-2.0 * resolution. The effect of estimated contact distance error is usually minimal as the thickness is small and the difference between thickness and particle spacing is also small. It does, however, have an effect in some cases. We will factor thickness into the contact distance in the near future.

Okay, understand. I'm not sure when we have time to test other cases yet. We'll try as soon as possible tho.

SPHINXsys/src/shared/particle_dynamics/solid_dynamics/contact_dynamics.cpp
```diff
@@ -129,6 +126,27 @@ namespace SPH
...
+			for (size_t k = 0; k < contact_configuration_.size(); ++k)
+			{
+				StdLargeVec<Real>& contact_Vol_k = *(contact_Vol_[k]);
+				Neighborhood& contact_neighborhood = (*contact_configuration_[k])[index_i];
+				for (size_t n = 0; n != contact_neighborhood.current_size_; ++n)
+				{
+					Real corrected_W_ij = std::max(contact_neighborhood.W_ij_[n] - offset_W_ij_[k], 0.0);
+					sigma += corrected_W_ij * contact_Vol_k[contact_neighborhood.j_[n]];
+				}
+				constexpr Real heuristic_limiter = 0.005;
```
FabienPean-Virtonomy, line 144:
```suggestion
				constexpr Real heuristic_limiter = 0.005;
				// The maximum contact pressure is the bulk modulus K=density*c*c,  wherein c is the sound wave speed. 
				// The contact pressure is huge and applied on fewer particles than for solids, yielding high acceleration locally, which is one source of instability. 
```

FabienPean-Virtonomy: Wouldn't it be good to look at whether this value can be replaced by the shell thickness since it seems to be in the same range ?

BenceVirtonomy: I think the thickness should not only be a multiplier but the dp be replaced by the thickness for calculating contact distance and thus the contact density

DongWuTUM: Yes, the dp should be replaced by thickness in some way. It is important that we need to ensure that both bodies can search for each other's particles as their neighbor particles when updating `contact_configuration_`.

FabienPean-Virtonomy (approved):

PR #203, opened 2023-02-15
Xiangyu/using data index

Xiangyu-Hu:
Please have a check on this revision in which both 2d and 3d vectors or matrices are allow to be particle variables.

### DIFF
--- a/SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.cpp
+++ b/SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.cpp
@@ -20,6 +20,81 @@ namespace SPH
 		father_element.insertNodeAfter(father_element.node_end(), SimTK::Xml::Element(child_name));
 	}
 	//=================================================================================================//
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite,
+		const std::string& attrib_name, const Real& value)
+	{
+		SimTK::Xml::Attribute attr_(attrib_name, SimTK::String(value));
+		ele_ite->setAttributeValue(attr_.getName(), attr_.getValue());
+	}
+	//=================================================================================================//
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite, 
+		const std::string& attrib_name, const int& value)
+	{
+		SimTK::Xml::Attribute attr_(attrib_name, SimTK::String(value));
+		ele_ite->setAttributeValue(attr_.getName(), attr_.getValue());
+	}
+	//=================================================================================================//
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite, 
+		const std::string& attrib_name, const Vec2d& value)
+	{
+		setAttributeToVectorElement<2>(ele_ite, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite,
+		const std::string& attrib_name, const Vec3d& value)
+	{
+		setAttributeToVectorElement<3>(ele_ite, attrib_name, value);
+	}
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite, 
+		const std::string& attrib_name, const Mat2d& value)
+	{
+		setAttributeToMatrixElement<2>(ele_ite, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite, 
+		const std::string& attrib_name, const Mat3d& value)
+	{
+		setAttributeToMatrixElement<3>(ele_ite, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Real& value)
+	{
+		std::string value_in_string = ele_ite_->getRequiredAttributeValue(attrib_name);
+		value = SimTK::convertStringTo<Real>(value_in_string);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, int& value)
+	{
+		std::string value_in_string = ele_ite_->getRequiredAttributeValue(attrib_name);
+		value = SimTK::convertStringTo<int>(value_in_string);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Vec2d& value)
+	{
+		getVectorAttributeValue<2>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Vec3d& value)
+	{
+		getVectorAttributeValue<3>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Mat2d& value)
+	{
+		getMatrixAttributeValue<2>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Mat3d& value)
+	{
+		getMatrixAttributeValue<3>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
 	void XmlEngine::writeToXmlFile(const std::string &filefullpath)
 	{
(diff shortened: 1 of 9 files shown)


SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.h
```diff
@@ -69,100 +68,72 @@ namespace SPH
 		void addChildToElement(SimTK::Xml::Element &father_element, const std::string &child_name);
 
 		/** Add an attribute of type string to an xml element.  */
-		template <class T>
-		void setAttributeToElement(const SimTK::Xml::element_iterator &ele_ite, const std::string &attrib_name, const T &value);
-
-		void setAttributeToElement(const SimTK::Xml::element_iterator &ele_ite, const std::string &attrib_name, const int &value)
+		void setAttributeToElement(const SimTK::Xml::element_iterator &ele_ite, const std::string &attrib_name, const Real &value)
```
RusellSyne, line 71: I think the modification of the member function "SPH::XmlEngine::setAttributeToElement" will cause some issues.

For the last input parameter of the member function "SPH::XmlEngine::setAttributeToElement", I think several overloaded versions are missed. For example, when the data type of the last input parameter is "Vecd" or "_Ty", some error occurs like this:

image

RusellSyne (changes requested):

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.h
RusellSyne: It seems that this issue during compiling on occurs for Windows.  Linux doesn't have this issue and the cases can pass the ctest. Then, I will test if I can define a 3d vector/matrix in 2D cases and also 2d vector/matrix in 3D cases in Linux.

RusellSyne:
**On Linux system**, the definition and usage of 3D vector/matrix in 2D cases (and 2D vector/matrix in 3D cases) have been **tested and passed**. While on **Windows**, there are some **issues during compiling** as shown above.

A further issue may relate to this:
When we use "addBodyStateForRecording" to output data, only 2D vector/matrix are allowed to output in 2D cases while 3D vector/matrix are not; also in 3D cases, only 3D vector/matrix are allowed to output while 2D vector/matrix are not. 
If we use "addBodyStateForRecording" to output 3D vector/matrix in 2D cases, although no bugs/warnings occur, the 3D vector/matrix are not outputted, and the same situation for 3D cases.

FabienPean-Virtonomy:
I could reproduce the failure to compile on Windows in a minimal snippet: https://godbolt.org/z/6Khq6joeY
This seems to be a recurring problem in MSVC that does not handle well template argument deduction with default and/or self-referencing template arguments [[1]](https://stackoverflow.com/questions/58520935/avoiding-ambiguous-parameters-in-templated-eigen-functions-with-msvs)

Ways around that are given in the godbolt link: either add a variadic template parameters (function bar) to fully capture the Eigen parameters or make a catch-all Eigen argument (https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html) that can be dispatched with `if constexpr` (function babar).

Besides that, @RusellSyne raise a good point on the `addBodyStateForRecording`. What is the use-case encountered that push for this change ?

RusellSyne:
Hi @FabienPean-Virtonomy. About `addBodyStateForRecording`, the current cases have no need to output 3D Vector/Matrix in 2D cases (and 2D Vector/Matrix in 3D cases), but I need this feature in the following cases. And also, when I tested the definition and usage of 3D Vector/Matrix in 2D cases, I want to output 3D Vector/Matrix data for verification, which I found was not allowed in 2D cases (the same situation in 3D cases).

Xiangyu-Hu:
> I could reproduce the failure to compile on Windows in a minimal snippet: https://godbolt.org/z/6Khq6joeY This seems to be a recurring problem in MSVC that does not handle well template argument deduction with default and/or self-referencing template arguments https://stackoverflow.com/questions/58520935/avoiding-ambiguous-parameters-in-templated-eigen-functions-with-msvs
> 
> Ways around that are given in the godbolt link: either add a variadic template parameters (function bar) to fully capture the Eigen parameters or make a catch-all Eigen argument (https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html) that can be dispatched with `if constexpr` (function babar).
> 
> Besides that, @RusellSyne raise a good point on the `addBodyStateForRecording`. What is the use-case encountered that push for this change ?

I solved this issue by using a template function within "SPH::XmlEngine::setAttributeToElement".

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.h
```diff
@@ -92,28 +96,30 @@ namespace SPH
...
+		void getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, const std::string& attrib_name, Vec3d& value);
+
 		template <int DIMENSION>
-		void getRequiredAttributeValue(SimTK::Xml::element_iterator &ele_ite_, const std::string &attrib_name,
-									   Eigen::Matrix<Real, DIMENSION, DIMENSION> &value)
+		void getMatrixAttributeValue(SimTK::Xml::element_iterator& ele_ite_, const std::string& attrib_name,
+			Eigen::Matrix<Real, DIMENSION, DIMENSION>& value)
 		{
 			std::string value_in_string = ele_ite_->getRequiredAttributeValue(attrib_name);
 			SimTK::Array_<Real, int> array_;
```
FabienPean-Virtonomy, line 125: Why is this done only for matrices and not vectors ?

Both should be serializable uniformly according to [SimTK::String(const T&)](https://simtk.org/api_docs/simbody/api_docs33/Simbody/html/classSimTK_1_1String.html#ae1c3122a261b2a809607040facdd7c07) and [operator<< for SimTK::Mat](https://simtk.org/api_docs/simbody/api_docs33/Simbody/html/namespaceSimTK.html#abf8a923ce5ce85ac3fddbddd1c5df70d). This would reduce duplication and simplify the code to do the same for vectors and matrices.

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.h
```diff
@@ -67,22 +67,26 @@ namespace SPH
...
+		void setAttributeToElement(const SimTK::Xml::element_iterator& ele_ite, const std::string& attrib_name, const Vec3d& value);
+
 		template <int DIMENSION>
-		void setAttributeToElement(const SimTK::Xml::element_iterator &ele_ite, const std::string &attrib_name,
-								   const Eigen::Matrix<Real, DIMENSION, DIMENSION> &value)
+		void setAttributeToMatrixElement(const SimTK::Xml::element_iterator& ele_ite, const std::string& attrib_name,
+										 const Eigen::Matrix<Real, DIMENSION, DIMENSION>& value)
 		{
 			SimTK::Array_<Real, int> array_(DIMENSION * DIMENSION);
 			for (int i = 0; i < DIMENSION; i++)
```
FabienPean-Virtonomy, line 92: Why is this done only for matrices and not vectors ?

Both should be serializable uniformly according to [SimTK::String(const T&)](https://simtk.org/api_docs/simbody/api_docs33/Simbody/html/classSimTK_1_1String.html#ae1c3122a261b2a809607040facdd7c07) and [operator<< for SimTK::Mat](https://simtk.org/api_docs/simbody/api_docs33/Simbody/html/namespaceSimTK.html#abf8a923ce5ce85ac3fddbddd1c5df70d). This would reduce duplication and simplify the code to do the same for vectors and matrices.

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.cpp
```diff
@@ -20,6 +20,81 @@ namespace SPH
...
+	{
+		getMatrixAttributeValue<2>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
+	void XmlEngine::getRequiredAttributeValue(SimTK::Xml::element_iterator& ele_ite_, 
+		const std::string& attrib_name, Mat3d& value)
+	{
+		getMatrixAttributeValue<3>(ele_ite_, attrib_name, value);
+	}
+	//=================================================================================================//
```
FabienPean-Virtonomy, line 97: Too much duplication

FabienPean-Virtonomy (commented):
The last commit (and existing differentiated code path for Vec and Mat) generates duplication and boilerplate which would likely be overcome with generic programming.

Consider more closely using
```cpp
template<int Dim, auto... Args>
void v1(const Eigen::Matrix<double,Dim,1,Args...>& a) 
{}
template<int Dim, auto... Args>
void v1(const Eigen::Matrix<double,Dim,Dim,Args...>& a)
{}
```
 or

```cpp
template<int NRows, int NCols,, auto... Args>
void v2(const Eigen::Matrix<double,NRows,NCols,Args...>& a) 
{}
```

or 

```cpp
template<typename Derived>
void v3(const Eigen::MatrixBase<Derived>& a)
{}
```

RussellSyne (approved):

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.cpp
Xiangyu-Hu: Yes. However, I am not good enough to decrease the repeating. 
I will try to revise it later.

SPHINXsys/src/shared/simbody_sphinxsys/xml_engine.h
Xiangyu-Hu: Another reason the String function is for simtk and our data is from eigen3. 
This leads to some issue.
However, any better revision is welcomed.

PR #210, opened 2023-02-20
Hotfix/clarify muscle constitute relarion

Xiangyu-Hu:
Hi Benjemin,
I have find that the J in muscle equation is OK as is from the reference paper listed in Neohook model.
Xiangyu


JohnVirtonomy:
The J is generally ok. But haven't you already multiplied by J?

image

Xiangyu-Hu:
> The J is generally ok. But haven't you already multiplied by J?
> 
> image

With the energy from Eq. (13) of the reference paper, you will the the extra J is actually correct.

JohnVirtonomy:
> With the energy from Eq. (13) of the reference paper, you will the the extra J is actually correct.

yes, that's right. If you replace lnJ by (j-1) before derivation, the extra J is fine.

JohnVirtonomy (approved):

PR #223, opened 2023-03-19
Xiangyu/material local parameters

Xiangyu-Hu:
the local material parameters now are reloaded together with particle reloading.
For this, particle generators are slightly modified.
After this, no XML engine needed in material classes and reloading are generalized.


SPHINXsys/src/shared/particle_generator/base_particle_generator.cpp
```diff
@@ -72,4 +73,12 @@ namespace SPH
 		base_particles_.readFromXmlForReloadParticle(file_path_);
 	}
 	//=================================================================================================//
+	void ParticleGeneratorReload::generateParticlesWithBasicVariables()
+	{
+		base_material_.registerReloadLocalParameters(&base_particles_);
+		initializeGeometricVariables();
+		// should be determined first before register other variables
+		base_particles_.real_particles_bound_ = base_particles_.total_real_particles_;
+	}
```
FabienPean-Virtonomy, line 82: Is the lack of symmetry with lines 22-28 normal ?

```cpp
	void BaseParticleGenerator::generateParticlesWithBasicVariables()
	{
		initializeGeometricVariables();
		// should be determined first before register other variables
		base_particles_.real_particles_bound_ = base_particles_.total_real_particles_;
		base_material_.registerReloadLocalParameters(&base_particles_);
	}
```

FabienPean-Virtonomy (approved):

SPHINXsys/src/shared/particle_generator/base_particle_generator.cpp
Xiangyu-Hu: this due to the different initialization mechanism of particle reload, in which when the local parameters variable is registered, there is no data yet.

DongWuTUM (approved):

PR #224, opened 2023-03-21
Fix/integer power

Xiangyu-Hu:
1. revise the usage of pow() so that https://en.cppreference.com/w/cpp/language/unqualified_lookup can be implemented later. 
2.  correct a bug in regression test and revise the regression data base also,


SPHINXsys/src/shared/adaptations/adaptation.cpp
```diff
@@ -90,7 +90,7 @@ namespace SPH
 	{
 		// estimate the required mesh levels
 		size_t total_levels = (int)log10(MinimumDimension(shape.getBounds()) / ReferenceSpacing()) + 2;
```
FabienPean-Virtonomy, line 92:
```suggestion
		int total_levels = (int)log10(MinimumDimension(shape.getBounds()) / ReferenceSpacing()) + 2;
```

SPHINXsys/src/shared/adaptations/adaptation.cpp
```diff
@@ -90,7 +90,7 @@ namespace SPH
...
-		Real coarsest_spacing = ReferenceSpacing() * powerN(2.0, total_levels - 1);
+		Real coarsest_spacing = ReferenceSpacing() * pow(2.0, (int)total_levels - 1);
```
FabienPean-Virtonomy, line 93:
```suggestion
		Real coarsest_spacing = ReferenceSpacing() * pow(2.0, total_levels - 1);
```

tests/3d_examples/test_3d_twisting_column/regression_test_tool/MyObserver_Position_Run_0_result.xml
FabienPean-Virtonomy, line 1: Why are regression test files updated ?

FabienPean-Virtonomy (approved):

PR #225, opened 2023-03-23
add test_2d_dambreak_python case with pybind

maiyetum95:


Xiangyu-Hu:
> _ No description provided. _

Please provide a description to the changes.

tests/2d_examples/test_2d_dambreak_python/pybind_tool/pybind_test.py
maiyetum95, line 1: Add the path of the generated files from cpp (shared object in LInux and dynamic link libraries in Windows) to the python script, which should be corresponds to the folder where the file is generated.

tests/2d_examples/test_2d_dambreak_python/CMakeLists.txt
maiyetum95, line 1: If there are several python version in the computer, make sure that the python version here running the python script should be the same with the version which generated dynamic files.

.github/workflows/ci.yml
maiyetum95, line 1: Considering that the pybind is used as a submodule, we should initialize and update the submodule.

CMakeLists.txt
maiyetum95, line 1: In the main CMakeLists.txt, we add Position Independent Code (PIC) for Linux compile. 
Pybind only have head files, so we use add_subdirectory to add the module.

SPHINXsys/src/for_2D_build/geometries/level_set_supplementary.cpp
maiyetum95, line 1: Fix the problem when use Linux to compile the shared objects.

tests/2d_examples/test_2d_dambreak_python/pybind_tool/pybind_tool.py
maiyetum95, line 1: use python to run the 2d dambreak case, there are two arguments can be changed. The first is for setting restart step, the second is the total running time for the case.

.github/workflows/ci.yml
maiyetum95: This pybind11_dir is not necessary if use add subdirectory in cmakelists

tests/2d_examples/test_2d_dambreak_python/dambreak_python.cpp
maiyetum95, line 1: didn't change the main part when compared with the original dambreak case, only add cmakeTest() function for ctest when call the pybind_test.py

Xiangyu-Hu:
Its test finished in 0.11 second? It is strange.

tests/2d_examples/test_2d_dambreak_python/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+     DESTINATION ${BUILD_BIND_PATH})
+
+aux_source_directory(. DIR_SRCS)
+pybind11_add_module(${PROJECT_NAME} ${EXECUTABLE_OUTPUT_PATH} ${DIR_SRCS} )
+set_target_properties(${PROJECT_NAME} PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY "${EXECUTABLE_OUTPUT_PATH}")
+target_link_libraries(${PROJECT_NAME} PRIVATE sphinxsys_2d)
+
+if(${CMAKE_SYSTEM_NAME} MATCHES Windows)
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+    set(PYTHON_EXECUTABLE "C:/hostedtoolcache/windows/Python/3.9.13/x64/python3.exe")
```
FabienPean-Virtonomy, line 29: This is set automatically when the Python dependency is correctly in the CMake files, with `find_package(Python ...)`, see https://cmake.org/cmake/help/latest/module/FindPython.html

FabienPean-Virtonomy (changes requested):
Please write a description and reasons to the PR by editing your main comment in the conversation.

PR #227, opened 2023-03-24
Heaviside

Xiangyu-Hu:
using more accurate volume fraction function in level set.
surface bounding introduced in static confinement to eliminate the near surface artifact.


SPHINXsys/src/shared/geometries/level_set.cpp
```diff
@@ -17,17 +17,25 @@ namespace SPH
...
+	Real BaseLevelSet::Heaviside(Real phi, Real half_width)
 	{
 		Real heaviside = 0.0;
 		Real normalized_phi = phi / half_width;
 		if (phi < half_width && phi > -half_width)
-			heaviside = (0.5 + 0.5 * normalized_phi) + 0.5 * sin(Pi * normalized_phi) / Pi;
+			heaviside = 0.5 + 0.5 * normalized_phi;
 		if (normalized_phi > 1.0)
 			heaviside = 1.0;
 		return heaviside;
```
FabienPean-Virtonomy, line 28:
```suggestion
		Real normalized_phi = phi / half_width;
		return std::clamp(0.5 + 0.5 * normalized_phi, 0.0, 1.0);
```

SPHINXsys/src/shared/geometries/level_set.cpp
```diff
@@ -17,17 +17,25 @@ namespace SPH
...
 	}
 	//=================================================================================================//
+	Real BaseLevelSet::computeVolumeFraction(Real cell_center_phi_, Real data_spacing, const Vecd position)
+	{
+		Vecd normal_direction = probeNormalDirection(position);
+		normal_direction /= normal_direction.norm() + TinyReal;
+		Real factor = 0.5 / normal_direction.cwiseAbs().maxCoeff();
+		return Heaviside(cell_center_phi_, factor * data_spacing);
```
FabienPean-Virtonomy, line 36: Calculation seems weird. If we take a normal along the cell diagonal [sqrt(2)/2,sqrt(2)/2] this does not work really well.

FabienPean-Virtonomy (changes requested):

SPHINXsys/src/for_2D_build/geometries/level_set_supplementary.cpp
```diff
@@ -427,7 +429,8 @@ namespace SPH
 	Vecd LevelSet::computeKernelGradientIntegral(const Vecd &position)
 	{
 		Real phi = probeSignedDistance(position);
-		Real cutoff_radius = kernel_.CutOffRadius(global_h_ratio_);
+		Real corrected_global_h_ratio = global_h_ratio_ * 1.01;
```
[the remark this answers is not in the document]
Xiangyu-Hu: Sorry. This is not the final version, in which the change is deleted.

SPHINXsys/src/shared/geometries/level_set.cpp
Xiangyu-Hu: The function is moved to scalar_function.cpp as you suggested.

SPHINXsys/src/shared/geometries/level_set.cpp
Xiangyu-Hu: I basically want to take the maximum absolute value of the vector. Do you have a proper way?

Xiangyu-Hu: Yes. We have developed the simple method which would yield the same result as we have introduced the contribution of normal direction. The idea is first determining the direction of plane them computing the cut.

FabienPean-Virtonomy: The simplified method yields quite different results from analytical. Why not simply computing the exact value when the plane actually cut the cell ?

FabienPean-Virtonomy (approved):

SPHINXsys/src/shared/geometries/level_set.cpp
Xiangyu-Hu: I have updated a more accurate version. In case you have even more accurate one, you can replace it.

PR #228, opened 2023-03-31
Poiseuille flow with shell failing

YuVirtonomy:
shell problem 

**_Thickness of shell_**
its seem there is some problem with **shell_thickness = resolution_ref * 1.0,**. In this condition, the shell wall is too weak to prevent fluid particles from penetrating the shell wall. 

With **shell_thickness = resolution_ref * 2.0,** , the result seem more make sense And, shell can prevent the fluid from penetrating

The following result used shell_thickness = resolution_ref * 1.0;
image

The following results used  shell_thickness = resolution_ref * 2.0;
image


YuVirtonomy:
image

Xiangyu-Hu:
It seems that there is no test for the cases.

YuVirtonomy:
> It seems that there is no test for the cases.

The test case is in
tests/3d_examples/test_3d_poiseuille_flow_shell/poiseuille_flow_shell.cpp

Xiangyu-Hu:
Sorry. I mean not Google test or regression test.

PR #233, opened 2023-04-14
change the Eulerian cases

ZhentongWang:
change the Euelrian cases


Xiangyu-Hu:
> change the Euelrian cases

Zhengtong, The code is not able to built in linux and mcos.

Xiangyu-Hu:
@ZhentongWang the 2d flow around cylinder have not run in the Ctest.

PR #239, opened 2023-04-21
Fix index error in `ShellContactDensity`

FabienPean-Virtonomy:


BenceVirtonomy (approved):

Xiangyu-Hu:
@DongWuTUM Hi, Dong. Could you check whether this bug fix is ok? It failed a case in Mc OS.

DongWuTUM:
> @DongWuTUM Hi, Dong. Could you check whether this bug fix is ok? It failed a case in Mc OS.

OK, I will check it.

FabienPean-Virtonomy:
@DongWuTUM The regression data were generated with the bug present at https://github.com/Xiangyu-Hu/SPHinXsys/commit/ae66a376fc9807033dd63836fbb46bfe35143339

Xiangyu-Hu:
Thanks for the revision.

FabienPean-Virtonomy:
I don't know why macOS is hanging. Tried, several variations, if you have one feel free to fix.

DongWuTUM:
@FabienPean-Virtonomy Hi Fabien. Thank you very much for your revision. Seems like the problem is not located in the regression data. I will further check it.

Xiangyu-Hu:
@ncsalis could you have a look on the branch at mcos?

DongWuTUM:
The simulation sometimes crashed (the time step decreases to very small). This should be the problem.

BenceVirtonomy:
> The simulation sometimes crashed (the time step decreases to very small). This should be the problem. I will correct it.

@DongWuTUM we always check for time step decrease on our side, and stop the simulation if that happens.  It's easier to debug, I would recommend adding it to all simulations in general.

DongWuTUM:
> > The simulation sometimes crashed (the time step decreases to very small). This should be the problem. I will correct it.
> 
> @DongWuTUM we always check for time step decrease on our side, and stop the simulation if that happens. It's easier to debug, I would recommend adding it to all simulations in general.

@BenceVirtonomy Good idea! It should be so.

ncsalis:
> The simulation sometimes crashed (the time step decreases to very small). This should be the problem. I will correct it.

Same thing happens to me. Sometimes crashes with very low dt and sometimes is fine

DongWuTUM:
With adjusted limiter, all are Okay now.

PR #244, opened 2023-04-24
Cleanup CMake python targets

FabienPean-Virtonomy:


Xiangyu-Hu:
@maiyetum95 could you have check on the changes and approve it if is OK.

maiyetum95 (approved):
I think the current changes are appropriate.

PR #247, opened 2023-04-27
Add getters for base particle member variables

FabienPean-Virtonomy:
* Add getter for accessing reload and restart variable lists from base particles. This allows to externalize and add extra reload/restart writer/reader
* Change arguments into const


Xiangyu-Hu:
> * Add getter for accessing reload and restart variable lists from base particles. This allows to externalize and add extra reload/restart writer/reader
> * Change arguments into const

Do you have a test for this change?

FabienPean-Virtonomy:
A test for a const getter ??

Xiangyu-Hu (approved):

PR #274, opened 2023-05-11
Fix/clarify number density

Xiangyu-Hu:
1. Naming on number density revised better meaning.
2. @ncsalis  the free surface lifting issue is solved by using simpler density reinitialization without taking account previous step density value.


Xiangyu-Hu:
@ncsalis please review and approve the pull request.

ncsalis (approved):
pull274

Xiangyu-Hu:
@ncsalis Thanks for the clarification figure.

Bo-Zhang1995:
This pull request leads to the pressure fluctuation in the OWSC case.

Xiangyu-Hu:
> This pull request leads to the pressure fluctuation in the OWSC case.

According to Nicolo's study that the new free surface density estimation (simply choose from summation density of reference density) is more stable for the the test cases. 
It actually gives a simple cavitation model which senses the tension near structure wall. 
Therefore, I prefer to use new formulation. We can just report the fluctuative values or smooth it with explicit  mentioning the smoothing. Later, we can consider a sensing model so that  the resolution of sensing is considered.

Bo-Zhang1995:
This formulation will lead to another problem for the OWSC problem; as shown in the Figure, there will be an extreme peak value at the high slamming pressure. (The previous results using the half acoustic time step by mistake avoid this observation). Actually, this phenomenon also exists in the 2d OWSC. No related to the correction still exists in NKGC.
Would you happen to have any suggestions?
pressure

Bo-Zhang1995:
[    this->rho_[index_i] = SMAX(this->rho_sum_[index_i], this->rho0_);](https://github.com/Xiangyu-Hu/SPHinXsys/blob/d7c33a4ab0bd9e135dc22a3c4cf1bfe048b8ff16/src/shared/particle_dynamics/fluid_dynamics/density_summation.hpp)
Volume updates here may not be conducted here. Because we have moved the volume update to the density summation, but as I modified last time, not to modify the free surface part? I will try this first.

Bo-Zhang1995:
After fixing this bug,
1. the extreme peak pressure still exists. Other attempts should be made.
2. I have checked that the current density reinitialization formulation should be better than the previous one (regarding the wave height) except for the pressure fluctuation. Keep using the new formulation for density summation.
3. Chi's OWSC uses the old formulation, but his paper indicates the currently used density formulation.

Xiangyu-Hu:
> This formulation will lead to another problem for the OWSC problem; as shown in the Figure, there will be an extreme peak value at the high slamming pressure. (The previous results using the half acoustic time step by mistake avoid this observation). Actually, this phenomenon also exists in the 2d OWSC. No related to the correction still exists in NKGC. Would you happen to have any suggestions? pressure

I will simple keep it and explain how it can be made disappeared.

Xiangyu-Hu:
> > This formulation will lead to another problem for the OWSC problem; as shown in the Figure, there will be an extreme peak value at the high slamming pressure. (The previous results using the half acoustic time step by mistake avoid this observation). Actually, this phenomenon also exists in the 2d OWSC. No related to the correction still exists in NKGC. Would you happen to have any suggestions? pressure
> 
> I will simple keep it and explain how it can be made disappeared.

Do you mean this peak will show up with the old density formulation and full time step size?

PR #281, opened 2023-05-14
Neumann boundary condition test case

ChenxiZhaoTUM:
This is the first case to implement multiple boundary conditions, ie. Dirichlet and Neumann.
The case works on the new versions of base_local_dynamics and new particle_dynamics_diffusion_reaction.
Robin boundary has been deleted temporarily due to the number of arguments of the constructor (how to get the temperature at infinity while achieving diffusion relaxation process with Robin boundary).
Besides, the InnerRelation in RK2 class in cpp. files have been changed to BaseInnerRelation to unify the form with pkj_lv_electrocontraction case which calls for BaseContactRelation (electro_physiology.h:214).


Xiangyu-Hu:
@ChenxiZhaoTUM you need merge master to your branch first, as the master has changed the directory structure.
There should be should be no conflict for it.

PR #290, opened 2023-05-18
add aortic_valve.cpp

yuanjiajy:


Xiangyu-Hu:
@yuanjiajy you do not need to close the pull request, just transfer it draft request.

Xiangyu-Hu:
CI need pass before merge.

Xiangyu-Hu:
You are supposed not change the library directly.

Xiangyu-Hu:
I guess the problem is due to  the library libx11-dev and freeglut3-dev, may be some others, should be install first.

CMakeLists.txt
```diff
@@ -63,6 +63,10 @@ target_compile_definitions(sphinxsys_core INTERFACE _SILENCE_CXX17_ITERATOR_BASE
 find_package(Eigen3 CONFIG REQUIRED)
 target_link_libraries(sphinxsys_core INTERFACE Eigen3::Eigen)
 
+# ## OCCT
+find_package(OpenCASCADE CONFIG REQUIRED)
+target_link_libraries(sphinxsys_core INTERFACE TKSTEP  )
```
FabienPean-Virtonomy, line 68: The module relying on occt should be independent so that the user can choose to enable or disable it with an option

src/shared/common/vector_functions.cpp
```diff
@@ -3,6 +3,21 @@
...
+     gp_Pnt EigenToOcct(const Vec3d &eigen_vector)
+    {
+    return gp_Pnt(eigen_vector[0], eigen_vector[1], eigen_vector[2]);
+    }
+   //=================================================================================================//
+   Vec3d OcctVecToEigen(const gp_Vec &occt_vector)
+   {
+    return Vec3d(occt_vector.X(), occt_vector.Y(), occt_vector.Z());
+   }
+	//=================================================================================================//
```
FabienPean-Virtonomy, line 20: Please move things out of `src/shared` to keep occt bounded out of the core lib at the moment

yuanjiajy:
> I guess the problem is due to the library libx11-dev and freeglut3-dev, may be some others, should be install first.

Yes and I have changed it

src/shared/common/vector_functions.cpp
yuanjiajy: > Please move things out of `src/shared` to keep occt bounded out of the core lib at the moment

Yes i have removed things to the extra source part

.github/workflows/ci.yml
```diff
@@ -38,7 +38,9 @@ jobs:
             pkg-config `# for installing libraries with vcpkg`\
             git \
             cmake \
-            ninja-build
+            libx11-dev \
+            freeglut3-dev \
```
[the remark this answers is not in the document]
yuanjiajy: Actually it doesn't work when i only install the libgl-dev, so i install libgl-dev and freeglut3-dev.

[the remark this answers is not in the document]
FabienPean-Virtonomy: Yes, it doesn't work because the whole of opencascade is installed without distinction, including optional features relying on freeglut. As mentioned in my previous comment, you would need to modularize the vcpkg port to avoid installing everything, or rely on CMake `FetchContent` mechanism to only build and link TKSTEP.

PR #294, opened 2023-05-21
simplify the construct parameter for simbody coupling

Xiangyu-Hu:
@ncsalis @SPHYaru I have simplified the constructor for  simbody coupling.


ncsalis:
For me modifications are ok.

ncsalis (approved):

PR #299, opened 2023-05-23
change the Eulerian classes out of the master and edit the cmakelist to make the Eulerian cases run in the ctest.

ZhentongWang:
change the regression files further to make the tests successful in CI/MacOS


ZhentongWang:
also add Yongchuan as a collaborator to review the pull request

Xiangyu-Hu:
What did you changed for McOS?

ZhentongWang:
> What did you changed for McOS?

At the beginning, fails in macOS show that case of flow around cylinder is unsuccessful because of regression files, so I change the regression file in the Eulerian flow around cylinder case, specifically I decrease the value in the "generateDatabase" class to generate new regression files and then use these to replace the original regression files and now Eulerian flow around cylinder case could run successfully.

Xiangyu-Hu:
I think Zhengtong means that he decreased threshold for all systems.

Xiangyu-Hu:
> when we generate the regression files, we would use function .generateDataBase(**5.0e-3**) in class named RegressionTestDynamicTimeWarping. At the first, in the Eulerian flow around cylinder case, I set value as **5.0e-3** to generate the regression files but failed in CI/macOS, then I decrease the value from **5.0e-3** to **1.0e-3** to generate the regression files, and then results could pass in the CI/macOS.

This is OK.

Xiangyu-Hu (approved):

PR #308, opened 2023-05-29
add examples with the Laguerre Gauss kernel

ZhentongWang:
add three examples with the Laguerre Gauss kernel including 2D Eulerian Taylor-Green vortex and 2D Eulerian flow around cylinder and 3D passive cantilever case


ChiZhangatTUM:
Your CFL number is adoptable?

ZhentongWang:
> Your CFL number is adoptable?

Right now, I add the numerical dissipation  from the Riemann solver, so it is okay that CFL =0.6 in the Eulerian case now but CFL =0.3 in the solid dynamic. and Chi, I simulate some results in the issue called "Include the work on Lagueree Gauss kernel test case" , over there,  I simulate the case without numerical dissipation from the Riemann solver, but CFL has to decrease. but I haven't checked or used the high order time integration yet.

ChiZhangatTUM (approved):

PR #310, opened 2023-06-01
add another two minor criterion for regression test.

Bo-Zhang1995:
add another two minor criterion.
1. when the time step changes a lot, the result length will change sharply, leading to a DTW matrix with a lot of zero element, because we use the windows size to constrain the searching. Now, when the data length variation is greater than 10%, it will be regarded as the time step size has some problems and shows the error.
2. add another checking for whose case not including the database, but they are trying to do the regression test based on the dynamic time warping strategy. Now, if there is no database in the source and you are trying to do the regression test, an error will be outputted.


Bo-Zhang1995:
I have another two commits but why they don't appear at this pull request?

PR #315, opened 2023-06-05
Regression test add new critation

Bo-Zhang1995:


Xiangyu-Hu:
@Bo-Zhang1995 the CI has not pass for this pull request.

Bo-Zhang1995:
the test_2d_dambreak_python always has some problems. I still need to find the reason.

Xiangyu-Hu (approved):

PR #317, opened 2023-06-05
remove fluid particles

Shuaihao-Zhang:
I have removed the `FluidParticles` and  `ViscoelasticFluidParticles` in `fluid_particles.cpp `. `CompressibleFluidParticles` still remains because not sure where it is used.

After removing `FluidParticles` and  `ViscoelasticFluidParticles`, some basic cases such `dambreak`, `elastic_gate`, and `collision_beam ` can be run correctly.  But there are some cases that cannot be run and no error message output.

(1) Those cases need RunParticleRelaxation (such as `test_2d_collision `and `test_2d_fsi2`) can not be run, no error message and not sure about the reason.

(2) Those cases with multiphase flow (such as `test_2d_square_droplet` and `test_2d_two_phase_dambreak`) can not be run, no error message. I guess the problem is here:
https://github.com/Xiangyu-Hu/SPHinXsys/blob/b3f62fa6dae3032a3ac7a96cd478170068a3a77b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_multi_phase.cpp#L22

https://github.com/Xiangyu-Hu/SPHinXsys/blob/b3f62fa6dae3032a3ac7a96cd478170068a3a77b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_multi_phase.hpp#L112

I guess I used a wrong way to get the `contact_fluids_`.


Xiangyu-Hu:
> I have removed the `FluidParticles` and `ViscoelasticFluidParticles` in `fluid_particles.cpp `. `CompressibleFluidParticles` still remains because not sure where it is used.
> 
> After removing `FluidParticles` and `ViscoelasticFluidParticles`, some basic cases such `dambreak`, `elastic_gate`, and `collision_beam ` can be run correctly. But there are some cases that cannot be run and no error message output.
> 
> (1) Those cases need RunParticleRelaxation (such as `test_2d_collision `and `test_2d_fsi2`) can not be run, no error message and not sure about the reason.


The case "test_2d_collision" seems not relevant to the fluid particles.
Does "no error message" mean it crashes?
> 
> (2) Those cases with multiphase flow (such as `test_2d_square_droplet` and `test_2d_two_phase_dambreak`) can not be run, no error message. I guess the problem is here:
> 
> https://github.com/Xiangyu-Hu/SPHinXsys/blob/b3f62fa6dae3032a3ac7a96cd478170068a3a77b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_multi_phase.cpp#L22
> 
> https://github.com/Xiangyu-Hu/SPHinXsys/blob/b3f62fa6dae3032a3ac7a96cd478170068a3a77b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_multi_phase.hpp#L112
> 
> I guess I used a wrong way to get the `contact_fluids_`.

It looks that this way to obtain fluid material is OK.

Shuaihao-Zhang:
> 
> The case "test_2d_collision" seems not relevant to the fluid particles. Does "no error message" mean it crashes?
> 
Yes, all the cases with `RunParticleRelaxation` crashed, even though solid cases are not related to fluid particles. I will check again today.

Shuaihao-Zhang:
> > > The case "test_2d_collision" seems not relevant to the fluid particles. Does "no error message" mean it crashes?
> > 
> > 
> > Yes, all the cases with `RunParticleRelaxation` crashed, even though solid cases are not related to fluid particles. I will check again today.

This has been fixed.

The reason is that I added an extra line of code when I move the variable `SurfaceIndicator` from `FluidParticles `to `BaseParticles`.
https://github.com/Xiangyu-Hu/SPHinXsys/blob/b3f62fa6dae3032a3ac7a96cd478170068a3a77b/src/shared/particles/base_particles.cpp#L32

Don't remember why I did this. After deleting this line, the solid cases can be run.

Shuaihao-Zhang:
> 
> (2) Those cases with multiphase flow (such as `test_2d_square_droplet` and `test_2d_two_phase_dambreak`) can not be run, no error message. I guess the problem is here:
> 

This issue has been solved by registering the variable `SurfaceIndicator` in the material rather than in the `BaseParticles`. Otherwise, there will be some memory issues. Not very sure about the reason, but it seems that it is not allowed to register any new variables in the `BaseParticles`.

Xiangyu-Hu:
> > (2) Those cases with multiphase flow (such as `test_2d_square_droplet` and `test_2d_two_phase_dambreak`) can not be run, no error message. I guess the problem is here:
> 
> This issue has been solved by registering the variable `SurfaceIndicator` in the material rather than in the `BaseParticles`. Otherwise, there will be some memory issues. Not very sure about the reason, but it seems that it is not allowed to register any new variables in the `BaseParticles`.

It is allowed. However, it seems that  base_particles is not a right place to register it.
Surface indicator seems a proper shared variable registered in base material.

Issue #319, opened 2023-06-06
Numerical dissipation (non-linear wave fsi)


ncsalis:
Results are significantly worsened when simulating the case with high particle resolution (dx=0.015 circa 8M particles). This did not happened before merge #274 (prevented free surface rising when simulating with low particle resolution). Maybe it's better to revert this change since also @Shuaihao-Zhang encountered problems after this.

![issue](https://github.com/Xiangyu-Hu/SPHinXsys/assets/56539341/913e1cc4-369f-4a94-bd01-2c7d911460b3)

Xiangyu-Hu:
It really increases the numerical dissipation of the simulation. We probably can decrease the numerical dissipation, as a first check. As the present limiting parameter is based on previous formulation.

ncsalis:
I will check the limiter in the following days. I have found this issue also in the 2d_nlwfsi case, the numerical dissipation is quite high even at low particle resolutions.

ncsalis:
#324 is ineffective, with lower limiter the free surface elevation does not improve. Also the dynamics  (wave phase) of the wave is significantly modified and this aspect is unchanged

![eta](https://github.com/Xiangyu-Hu/SPHinXsys/assets/56539341/2455456b-bff4-457b-a5b0-31532beff36a)

ncsalis:
In the previous simulations there was a typo in the wave gauge position (1 meter downstream compared to the exp). The dynamics of the wave is good. I have checked the simulation at different merge (#229, #271 and #274). The numerical dissipation was always present, just hidden by the typo. 

![eta](https://github.com/Xiangyu-Hu/SPHinXsys/assets/56539341/54c0a052-f51e-408f-839f-14c05bb666ed)

Xiangyu-Hu:
I made another small change, but the result is similar.

ncsalis:
Thank you for the suggestion on Yaru's paper. I also was thinking that first order consistency should improve the result

ncsalis:
The first order consistency correction works very well. Result is now marginally improved

![wave_correct](https://github.com/Xiangyu-Hu/SPHinXsys/assets/56539341/651be78d-be81-494b-8214-4198d97d97c1)

ncsalis:
With the KGC seems that there is low dissipation thus with higher wave height that leads to higher amplitude of the fsi. How increase dissipation? I have tried by modifying the alpha parameter in the WKGC formulation but it doesn't produce significant result.

[...]

PR #324, opened 2023-06-06
319 high numerical dissipation non linear wave fsi

Xiangyu-Hu:
Let the limiter in Riemann solver a tunable value for cases. So that we can apply different test value for less numerical dissipation and explore an optimal value for general cases.
#319


ChiZhangatTUM:
The CI bug is fixed. You can merge from master and then PR again.

Xiangyu-Hu:
I close this pull as it is found the issue is due to a wrong setup of simulation.

PR #326, opened 2023-06-07
Remove fluid particles -- done

Shuaihao-Zhang:
refer to #317 
Four files have been removed:
fluid_particles.h
fluid_particles.cpp
fluid_particles_variable.h
fluid_particles_variable.cpp


ChiZhangatTUM:
@Shuaihao-Zhang Shuaihao, would you please briefly present the idea underlying this modification?

Shuaihao-Zhang:
> @Shuaihao-Zhang Shuaihao, would you please briefly present the idea underlying this modification?

Sure. After discussing with Prof. Hu, we think it's better to register all the variables in the dynamics part rather than in the particles. We just need `BaseParticles`, and the `FluidParticles `and `SolidParticles ` are not needed.

Here is one of the reason why using `FluidParticles` and `SolidParticles` is not good:

`FluidParticles` can only be used in `fluid_dynamics`, and `SolidParticles` can only be used in `solid_dynamics`. But for some specific situations, we want to use the functions both in `fluid_dynamics` and `solid_dynamics`. Then I don't know which particle should I defined for the materials.  For example, when I simulate the case `colliding_rubber_ring `based on update Lagrange method, I need to use the `BaseIntegration1stHalf` in the `fluid_dynamics `and also the  `ContactForce`  in `contact_dynamics`.

Since there are too many variables in `SolidParticles`, so currently I just remove `FluidParticles` first.

Shuaihao-Zhang:
For the fluid variables, only the `pressure `is registered in the material `Fluid`. `SurfaceIndicator `is registered in the `BaseParticles`. `DensityChangeRate `and `DensitySummation `are registered at the `fluid_dynamics`.

Xiangyu-Hu:
@ZhentongWang Within this pull request, fluid particles has been removed. Now there are some conflict with your merged code on FVM formulation. Could you help to solve the conflict? Thanks.

Xiangyu-Hu (approved):

Issue #335, opened 2023-06-13
No nonRiemannsolver for compressible flow in Eulerian SPH scheme.


Bo-Zhang1995:
Currently, it seems that only the HLLC Riemann solver with and without a limiter for the compressible flow is in the Eulerian SPH scheme.
It should include a non-Riemann solver as well for better testing of more cases.

Xiangyu-Hu:
@ZhentongWang please consider this issue.

ZhentongWang:
Sure, I will add the non-Riemann solver for compressible framework soon.

Xiangyu-Hu:
@ZhentongWang I have created a branch for this issue. Please work on this branch directly.

Xiangyu-Hu:
the issue is close due to the merged pull request #337

PR #337, opened 2023-06-13
FVM parser 

ZhentongWang:
main changing are following parts: 1. add .hpp file and remove some function into there. 2. add the FVM parser code and add two FVM cases including compressible and weakly compressbile flow. 3, add no Riemann solover in compressible flow


Xiangyu-Hu:
@ZhentongWang please use the branch used in #335 for No Riemann Problem code.

Xiangyu-Hu (approved):

PR #338, opened 2023-06-13
Use `tbb::affinity_partitioner` according to documentation

FabienPean-Virtonomy:
Following example https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Bandwidth_and_Cache_Affinity_os.html


ChiZhangatTUM:
How about the improvement of this modification?

FabienPean-Virtonomy:
I assume close to nil. Previous version was creating an `affinity_partitioner` for each translation unit which is pretty close to this. This may have had an impact when the `affinity partitioner` object is used multiple times in different algorithms within the same source file. I just tried to follow the documentation for the recommended use of this feature.

If we trust solely the CI (which is not really representative of actual load because of using 2/3 cores only) then it seemingly gets worse for macOS based on a single sample

PR #347, opened 2023-06-22
change FVM work

ZhentongWang:
the change contains the following part: 1. combine FVM and Eulerian SPH. 2. add ghost particle. 3. change the acoustic timestep in FVM , now this timestep is calculated related to the maximun node distance.4. right now , the density and energy are updated one step in the densityrelaxation only.


Xiangyu-Hu:
@FengWang3119 Could you also have look on the FVM version of the code?

FengWang3119:
OK, I will spend time learning the FVM code

Xiangyu-Hu (approved):

PR #351, opened 2023-06-26
CI use single

ChiZhangatTUM:
add build with single float without test in CI.


ChiZhangatTUM:
The present code works fine with GNU 9.4 while fails CI which use GNU 11.3 when build with using float. I will have double check whether this is due to the difference of GNU version.

Xiangyu-Hu:
@ChiZhangatTUM single precision should be a separate action from other three auctions.

ChiZhangatTUM:
> @ChiZhangatTUM single precision should be a separate action from other three auctions.

Like this?

  Linux-float:
    runs-on: ubuntu-22.04

  Linux-double:
    runs-on: ubuntu-22.04

PR #364, opened 2023-07-07
New aortic valve

yuanjiajy:
New Aortic-valve.cpp


ChiZhangatTUM:
Looks like all fails related to the installation of opencascade. 
One small question, can we replace Simbody's geometry with related staff in opencascade?

yuanjiajy:
> Looks like all fails related to the installation of opencascade. One small question, can we replace Simbody's geometry with related staff in opencascade?

Yes perhaps cause the capitalisation of OCCT and I have changed it. From my point of view, it could be changed since the latter one has more geometry functions

ChiZhangatTUM:
opencascade is only supported on '!(uwp | osx | arm)', which does not match x64-osx. This usually means that there are known build failures, or runtime problems, when building other platforms. To ignore this and attempt to build opencascade anyway, rerun vcpkg with `--allow-unsupported`.
you may add `--allow-unsupported` to vcpkg for MAC CI to have a check.

yuanjiajy:
> opencascade is only supported on '!(uwp | osx | arm)', which does not match x64-osx. This usually means that there are known build failures, or runtime problems, when building other platforms. To ignore this and attempt to build opencascade anyway, rerun vcpkg with `--allow-unsupported`. you may add `--allow-unsupported` to vcpkg for MAC CI to have a check.

OK I will do it later

yuanjiajy:
In addition to the inability to install OCCT in Mac, there are still problems when installing OCCT under windows.

ChiZhangatTUM:
> In addition to the inability to install OCCT in Mac, there are still problems when installing OCCT under windows.

Seems they already fixed this error. 
https://github.com/microsoft/vcpkg/issues/27896

ChiZhangatTUM:
I have tried with Mac OS 13.4.1, and its seems that oencascade installation works fine with "vcpkg install --clean-after-build --allow-unsupported ", while the installation took a very very long time. 

Therefore, 
for MAC CI, 
you can try the following modification 

first try: 

${{github.workspace}}/vcpkg/vcpkg install --clean-after-build --allow-unsupported 

if not work, 
try 
runs-on: macos-13

(Ref. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

yuanjiajy:
> I have tried with Mac OS 13.4.1, and its seems that oencascade installation works fine with "vcpkg install --clean-after-build --allow-unsupported ", while the installation took a very very long time.
> 
> Therefore, for MAC CI, you can try the following modification
> 
> first try:
> 
> ${{github.workspace}}/vcpkg/vcpkg install --clean-after-build --allow-unsupported
> 
> if not work, try runs-on: macos-13
> 
> (Ref. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

OK I will have try it

ChiZhangatTUM:
Interesting,  Windows now take more time than  Linux and Mac, in the previous version, it takes more time than the others.

yuanjiajy:
> Interesting, Windows now take more time than Linux and Mac, in the previous version, it takes more time than the others.

Yes, I have blocked the OCCT installation and stoopped my case running in Windows because github doesn't have supplement memory to support windows OCCT installations.

Xiangyu-Hu:
@yuanjiajy You can change the draft pull request to formal one already.

yuanjiajy:
> @yuanjiajy You can change the draft pull request to formal one already.

yuanjiajy:
> @yuanjiajy You can change the draft pull request to formal one already

> @yuanjiajy You can change the draft pull request to formal one already.

ok I have changed it

FabienPean-Virtonomy:
Is this PR a duplicate of #290 ?

CMakeLists.txt
```diff
@@ -66,6 +66,14 @@ target_compile_definitions(sphinxsys_core INTERFACE _SILENCE_CXX17_ITERATOR_BASE
 find_package(Eigen3 CONFIG REQUIRED)
 target_link_libraries(sphinxsys_core INTERFACE Eigen3::Eigen)
 
+# ## OCCT
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+find_package(OpenCASCADE CONFIG REQUIRED)
+target_link_libraries(sphinxsys_core INTERFACE TKSTEP TKSTEPAttr TKSTEP209 TKSTEPBase TKXSBase TKShHealing TKTopAlgo TKGeomAlgo TKBRep TKGeomBase TKG3d TKG2d TKMath TKernel )
```
FabienPean-Virtonomy, line 72: It seems unlikely you need all those targets?

CMakeLists.txt
```diff
@@ -66,6 +66,14 @@ target_compile_definitions(sphinxsys_core INTERFACE _SILENCE_CXX17_ITERATOR_BASE
 find_package(Eigen3 CONFIG REQUIRED)
 target_link_libraries(sphinxsys_core INTERFACE Eigen3::Eigen)
 
+# ## OCCT
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Windows")
```
FabienPean-Virtonomy, line 70: This add a lot of clutter throughout. Add the opencascade moduls under `modules/` folder instead to not have everything at root

opencascade_modules/aortic_valve_simulation/surface_shape.h
```diff
@@ -0,0 +1,67 @@
...
+#include "base_geometry.h"
+#include "simbody_middle.h"
+#include "opencascade/STEPCAFControl_Reader.hxx"
+#include "opencascade/XSControl_WorkSession.hxx"
+#include"opencascade/Transfer_TransientProcess.hxx"
+#include"opencascade/TopExp_Explorer.hxx"
+#include"opencascade/TopoDS.hxx"
+#include"opencascade/BRep_Builder.hxx"
+#include"opencascade/GeomAPI_ProjectPointOnSurf.hxx"
+#include"opencascade/ProjLib_ProjectOnSurface.hxx"
```
FabienPean-Virtonomy, line 25: You don't need all those headers here. Most should be in the `.cpp` You only need the header for `Standard_Real` and `Standard_CString`

CMakeLists.txt
yuanjiajy: > This add a lot of clutter throughout. Add the opencascade moduls under `modules/` folder instead to not have everything at root

Actually this part is used to stop OCCT installation in windows since Github hasn't enough storage to install OCCT

opencascade_modules/aortic_valve_simulation/surface_shape.h
yuanjiajy: > You don't need all those headers here. Most should be in the `.cpp` You only need the header for `Standard_Real` and `Standard_CString`

Yes perhaps I need to summarize them

Xiangyu-Hu:
@yuanjiajy you need resolve the conversations first, then I can merge it.

opencascade_modules/aortic_valve_simulation/surface_shape.h
yuanjiajy: > You don't need all those headers here. Most should be in the `.cpp` You only need the header for `Standard_Real` and `Standard_CString`

I have tried use them, But these two headers can't include the directory, and there is no generalised header that can include all the headers.

yuanjiajy:
> @yuanjiajy you need resolve the conversations first, then I can merge it.

I've tried the methods in the conversation

CMakeLists.txt
FabienPean-Virtonomy: Yes, I understand that from the condition. But you should not clutter the root CMakeLists with this stuff. You already have `opencascade_modules/CMakeLists.txt` where you can all OpenCascade related things.

opencascade_modules/aortic_valve_simulation/surface_shape.h
FabienPean-Virtonomy: The header file `surface_shape.h` only use the two typedef. You just need the OpenCascade header than define those. which seems to be `Standard_TypeDef.hxx`. The rest only needs to be in the `.cpp`

FabienPean-Virtonomy:
@Xiangyu-Hu OpenCASCADE can be added (or used without lengthy full build) with 3 main possibilities:
1. Modify https://github.com/microsoft/vcpkg/tree/master/ports/opencascade to make the vcpkg installation modular
2. Install via system package manager and use vcpkg [overlay ports](https://devblogs.microsoft.com/cppblog/using-system-package-manager-dependencies-with-vcpkg/) to find it without writing a dedicated CMake module file 
3. Use CMake `FetchContent` (see example below)

```cmake
# Options for libigl modules
option(LIBIGL_COPYLEFT_CGAL "Use CGAL" ON) # Enable desired dependency/component
option(LIBIGL_COPYLEFT_TETGEN "Use TetGen" OFF) # Disable undesired dependency/component
option(LIBIGL_USE_STATIC_LIBRARY "Use libigl as static library" OFF) # Disable undesired features/components
include(FetchContent)
FetchContent_Declare(
    libigl
    URL https://github.com/libigl/libigl/archive/refs/tags/v2.4.0.zip
    URL_HASH MD5=0b4fea5dba2117b8db85c99a39a71f83
)
FetchContent_MakeAvailable(libigl)
target_link_libraries(main PRIVATE igl::core)
```

opencascade_modules/aortic_valve_simulation/surface_shape.h
yuanjiajy: > The header file `surface_shape.h` only use the two typedef. You just need the OpenCascade header than define those. which seems to be `Standard_TypeDef.hxx`. The rest only needs to be in the `.cpp`

ok i will change this

FabienPean-Virtonomy: I just noticed below there is also `Handle_Geom_Surface` in the class, so you need the relevant header, likely something like `Handle_Geom_Surface.hxx`

yuanjiajy: > I just noticed below there is also `Handle_Geom_Surface` in the class, so you need the relevant header, likely something like `Handle_Geom_Surface.hxx`

ok i will change it

Xiangyu-Hu:
@yuanjiajy @FabienPean-Virtonomy  Have all issues settled? So that I can merge the pull request.

yuanjiajy:
> @yuanjiajy @FabienPean-Virtonomy Have all issues settled? So that I can merge the pull request.

For me, it's ok now

FabienPean-Virtonomy:
See #381 which is based on this work, and is an alternative way to organize the code and handle the integration of OpenCASCADE.

PR #366, opened 2023-07-10
[SYCL] Register device variables using data assemble

nR3D:
- Device memory is now allocated inside a data container assemble, as for other particle variables.
- Usage to `FluidParticles` has been changed to `BaseParticles`, and `getDeviceVariableByName` is used to pass address pointers to kernel classes.
- `DataTypeIndex` has been updated to include device types, which are only enabled if they are different from normal host types (using a new template parameter `Enable` for SFINAE instructions)
- `DeviceVariable` is a new data container type that manages device memory
- ~Variables that are only relevant to fluids (like Pressure, DensityChangeRate, and DensitySummation) are allocated with `BaseParticles::registerExtraDeviceMemory()`. This is a temporary solution; this task should be delegated to those classes that register their host counterparts, which should also be responsible to register the corresponding device variables.~ ([See comment below](https://github.com/Xiangyu-Hu/SPHinXsys/pull/366#issuecomment-1632950737))

Dambreak and Performance (benchmark) have been updated and can be run to test convergence and speedup, respectively.

### DIFF
--- a/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
+++ b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
@@ -328,7 +328,8 @@ public DeviceExecutable<BaseIntegration1stHalfWithWall<BaseIntegration1stHalfTyp
         : InteractionWithWall<BaseIntegration1stHalfType>(std::forward<Args>(args)...),
           DeviceExecutable<BaseIntegration1stHalfWithWall<BaseIntegration1stHalfType>,
                            BaseIntegration1stHalfWithWallKernel<typename BaseIntegration1stHalfType::DeviceKernel>>(this,
-                           *this->contact_configuration_device_, this->contact_particles_[0]->acc_device_,
+                           *this->contact_configuration_device_,
+                           this->contact_particles_[0]->template getDeviceVariableByName<DeviceVecd>("Acceleration"),
                            BaseIntegration1stHalfType::particles_,
                            BaseIntegration1stHalfType::inner_configuration_device_->data(),
                            this->riemann_solver_) {};
--- a/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
+++ b/src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
@@ -291,7 +291,7 @@ class BaseIntegration1stHalfWithWallKernel : public BaseIntegration1stHalfType {
                 RealT face_wall_external_acceleration = dot(acc_prior_i - acc_ave_k[index_j], -e_ij);
                 auto p_in_wall = p[index_i] + rho[index_i] * r_ij * SMAX(min_external_acc, face_wall_external_acceleration);
                 acceleration -= (p[index_i] + p_in_wall) * dW_ijV_j * e_ij;
-                rho_dissipation += riemann_solver.DissipativeUJump_Device(p[index_i] - p_in_wall) * dW_ijV_j;
+                rho_dissipation += riemann_solver.DissipativeUJump(p[index_i] - p_in_wall) * dW_ijV_j;
             }
         }
         acc[index_i] += acceleration / rho[index_i];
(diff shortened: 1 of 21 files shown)


src/shared/particles/base_particles.cpp
```diff
@@ -365,6 +367,61 @@ void BaseParticles::readFromXmlForReloadParticle(std::string &filefullpath)
     DataAssembleOperation<loopParticleVariables> loop_variable_namelist;
     loop_variable_namelist(all_particle_data_, variables_to_reload_, read_variable_from_xml);
 }
+
+    void BaseParticles::registerDeviceMemory() {
+        registerDeviceVariable<DeviceVecd>("Position", pos_.size(), pos_.data());
```
FabienPean-Virtonomy, line 372: You might want to use `total_real_particles_` instead on relying on vector size to avoid copy unnecessary buffer particles (unless particles switching and removal should also be performed on device)

nR3D: You are right, I can use `total_real_particles_`. I am wondering if those cases in which `total_real_particles_ != pos_.size()` would mean that `registerDeviceMemory` have been called at the wrong time, and if it should be handled with an error or a special behavior. Or it could also be considered as a precondition of `registerDeviceMemory` to have the right amount of particles.

FabienPean-Virtonomy: The main case when `total_real_particles_ != pos_.size()` is for fluid when there are inlet/outlet which require buffer particles (I don't know how the notion of ghost particle is used within SPHinXsys). Depending on the size of the domain this can make quite a difference.

nR3D:
> Variables that are only relevant to fluids (like Pressure, DensityChangeRate, and DensitySummation) are allocated with `BaseParticles::registerExtraDeviceMemory()`. This is a temporary solution; this task should be delegated to those classes that register their host counterparts, which should also be responsible to register the corresponding device variables.

@Xiangyu-Hu suggested to initialize those variables inside those kernels that are directly using them, like `BaseIntegration`. I have already pushed commit [c4814ee](https://github.com/Xiangyu-Hu/SPHinXsys/pull/366/commits/c4814eee7453907fe2399906ab690522d7885308) to this PR to address those changes.

Xiangyu-Hu:
Is this change in a new branch from master? Or you put it in you own sycl branch?

nR3D:
> Is this change in a new branch from master? Or you put it in you own sycl branch?

To my branch, which is already showing up in this PR, so that it will be merged together with the other changes.

> commit [c4814ee](https://github.com/Xiangyu-Hu/SPHinXsys/pull/366/commits/c4814eee7453907fe2399906ab690522d7885308)

src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
```diff
@@ -137,6 +141,30 @@ class BaseDensitySummationComplex
...
+    DensitySummationComplexKernel(const BaseDensitySummationComplexKernel& complexKernel,
+                                  const DensitySummationInnerKernel& innerKernel)
+        : BaseDensitySummationComplexKernel(complexKernel), DensitySummationInnerKernel(innerKernel) {}
+
+    template<class RealT, class InnerInteractionFunc, class ContactSummationFunc>
+    static void interaction(size_t index_i, Real dt, RealT* rho_sum, RealT rho0, RealT inv_sigma0, const RealT* mass,
+                            InnerInteractionFunc&& innerInteraction, ContactSummationFunc&& ContactSummation)
+    {
+        innerInteraction(index_i, dt);
+        RealT sigma = ContactSummation(index_i);
```
ChiZhangatTUM, line 155: "RealT" seems not good name. "RealType"?

ChiZhangatTUM:
The introduction of "DeviceReal", "DeviceVecd" and similar data seems confused for me. Would you please briefly explain the underlying consideration?

src/shared/particle_dynamics/fluid_dynamics/fluid_dynamics_complex.h
nR3D: Ok, I'll also change `Vec` to `VecType`, to keep the same nomenclature.

PR #369, opened 2023-07-15
Xiangyu/check fish swiming case

Xiangyu-Hu:
The problem of strange behavior for dual single time stepping seems due to the lack of numerical dissipation for momentum equation and the over correction of transport formulation for single time stepping.


SPHYaru (approved):

SPHYaru:
> The problem of strange behavior for dual single time stepping seems due to the lack of numerical dissipation for momentum equation and the over correction of transport formulation for single time stepping.

We added the numerical dissipation last week, and the problem still exists.   So next I try to decrease the correction of the transport formulation ?

Xiangyu-Hu:
> > The problem of strange behavior for dual single time stepping seems due to the lack of numerical dissipation for momentum equation and the over correction of transport formulation for single time stepping.
> 
> We added the numerical dissipation last week, and the problem still exists. So next I try to decrease the correction of the transport formulation ?

I leave tried with single time step with smaller coefficient for transport velocity correction. It works. The current transport velocity formulation is used for dual time criteria method..

PR #380, opened 2023-07-24
non-hourglass formulation for updated Lagrangian SPH

Shuaihao-Zhang:
The non-hourglass formulation for updated Lagrangian SPH.

Two basic cases are uploaded for testing: `test_2D_oscillating_beam` and `test_3D_oscillating_plate`. Regression tests are also included.


Xiangyu-Hu:
@Shuaihao-Zhang have you still kept the old method? It seems better to keep it.

Shuaihao-Zhang:
> @Shuaihao-Zhang have you still kept the old method? It seems better to keep it.

Yes, all the algorithms of old methods are still kept. I just don't provide a case for the old method, otherwise, there are too many cases. But it's easy to create a case based on the old algorithms.

Xiangyu-Hu (approved):

PR #381, opened 2023-07-25
Integration of OpenCASCADE for aortic valve particle generation

FabienPean-Virtonomy:
Alternative based upon initial work on #364


modules/opencascade/opencascade/relax_dynamics_surface.cpp
```diff
@@ -0,0 +1,108 @@
...
+            Standard_Real v;
+            gp_Pnt Point;
+            
+            pointonsurf.Parameters(1, u, v);
+
+            surface_shape_->surface_->D1(u, v, Point, tangent_u, tangent_v);
+            norm = tangent_u.Crossed(tangent_v);
+            normal_direction_ = OcctVecToEigen(norm);
+            n_[index_i] = normal_direction_.normalized();
+        }
```
EinSIMO, line 91: The variable name in this function should be modified. @yuanjiajy

modules/opencascade/opencascade/surface_shape.h
```diff
@@ -0,0 +1,59 @@
...
+#include <iostream>
+#include <string>
+#include <fstream>
+#include <filesystem>
+namespace fs = std::filesystem;
+
+namespace SPH
+{
+	
+
```
EinSIMO, line 31: remove the space.

modules/opencascade/opencascade/surface_shape.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return Vecd(actual_pnt[0], actual_pnt[1], actual_pnt[2]);
+}
+
+//=================================================================================================//
+SurfaceShapeSTEP::
+    SurfaceShapeSTEP(Standard_CString &filepathname, const std::string &shape_name)
+    : SurfaceShape(shape_name)
+{
+    STEPControl_Reader step_reader;
+    step_reader.ReadFile(filepathname);
```
EinSIMO, line 50: Check whether the file exist before readfile,  if not, report error message.

modules/opencascade/opencascade/surface_shape.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    return closest_pnt;
+}
+
+////=================================================================================================//
+bool SurfaceShape::checkContain(const Vecd &pnt, bool BOUNDARY_INCLUDED) { return 0; }
+BoundingBox SurfaceShape::findBounds() { return BoundingBox(); }
+//=================================================================================================//
+
+Vecd SurfaceShape::findActualPoint(Standard_Real u, Standard_Real v)
```
EinSIMO, line 33: findActualPoint, "Actual" seems confusing. Any better term?

modules/opencascade/opencascade/surface_shape.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+	//=================================================================================================//
+Vecd SurfaceShape::findClosestPoint(const Vecd &input_pnt)
+{
+
+    Vecd closest_pnt;
+    gp_Pnt point1 = EigenToOcct(input_pnt);
+    gp_Pnt point2;
+    Extrema_ExtAlgo Algo = Extrema_ExtAlgo_Tree;
+    point2 = GeomAPI_ProjectPointOnSurf(point1, surface_, Algo);
```
EinSIMO, line 20: seems line 18 and line 20 can be combined as gp_Pnt point2 = GeomAPI_ProjectPointOnSurf(point1, surface_, Algo);

modules/opencascade/opencascade/relax_dynamics_surface.cpp
```diff
@@ -0,0 +1,108 @@
...
+            surface_shape_(DynamicCast<SurfaceShape>(this, sph_body.body_shape_)), 
+            pos_(particles_->pos_), n_(*particles_->getVariableByName<Vecd>("NormalDirection")) {}
+
+        //=================================================================================================//
+        void SurfaceNormalDirection::update(size_t index_i, Real dt)
+        {
+            Extrema_ExtAlgo Algo = Extrema_ExtAlgo_Tree;
+            gp_Vec tangent_u, tangent_v;
+            gp_Vec norm;
+            Vecd normal_direction_;
```
EinSIMO, line 74: normal_direction_ change to normal_direction; "name_ " generally represents class member variable.

EinSIMO:
@FabienPean-Virtonomy Thanks for your help.

FabienPean-Virtonomy:
@EinSIMO My commits reorganized the physical layout of the opencascade integration. For all code related comments, I leave it up to @yuanjiajy to continue.

modules/opencascade/opencascade/surface_shape.cpp
FabienPean-Virtonomy: Also I would suggest to rename `find` too because the function only computes the cartesian coordinates from the parametric values. So something like `[compute or get][Cartesian or Spatial][Point or Coordinates or Coords]` might fit better

ChiZhangatTUM:
> @EinSIMO My commits reorganized the physical layout and the include blocks of the opencascade integration. For all code related comments, I leave it up to @yuanjiajy to continue.

@yuanjiajy Please revise the code accordingly.

yuanjiajy:
> Alternative based upon initial work on #364

hello  I want to solve the problems in the conversation but i have found the #381 is not based on the latest #364, for example: I have added opencascade.h to remove the opencascade in the header

yuanjiajy:
> @EinSIMO My commits reorganized the physical layout and the include blocks of the opencascade integration. For all code related comments, I leave it up to @yuanjiajy to continue.

Yes I will but it seems that it's not based in the latest #364

yuanjiajy:
> > @EinSIMO My commits reorganized the physical layout and the include blocks of the opencascade integration. For all code related comments, I leave it up to @yuanjiajy to continue.
> 
> @yuanjiajy Please revise the code accordingly.

ok I will check and find it.

yuanjiajy:
> @EinSIMO My commits reorganized the physical layout and the include blocks of the opencascade integration. For all code related comments, I leave it up to @yuanjiajy to continue.

Ok I will do that

modules/opencascade/opencascade/opencascade.h
FabienPean-Virtonomy, line 1: Why is this file back? I changed the header include blocks on purpose to reduce coupling.

yuanjiajy: > Why is this file back? I changed the header include blocks on purpose to reduce coupling.

I remember in the #364 , it is opencasade/....,hxx and i was suggested to remove the opencascade in the header, so I created  opencascade.h to inculude all the opencascade contents just like what we have done about Simbody. so the header in the surface.h is something like opencascade.h. If it's not good, I can change it.
f2610221f46aa11e6943f2e1c5e6fd0
f0180c967615240aba51ad16747c8cd

FabienPean-Virtonomy: Yes I suggested to remove it from the header to move them to the source files. I understand why you moved it into a new header now, but that's not the right answer to my comment then. I made it by design, so please revert those changes to how it was initially in this PR.

For further details: 
- Best practice is [include what you use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use). Yes catch-all headers can be convenient in some specific cases.
- Main objective is to encapsulate opencascade and minimize the public dependencies. By keeping most opencascade includes in the `.cpp` files, I could split public and private dependencies onto opencascade modules as such in the CMakeLists
```cmake
target_link_libraries(sphinxsys_opencascade PRIVATE TKSTEP TKSTEPAttr TKSTEP209 TKSTEPBase TKXSBase TKShHealing TKTopAlgo TKGeomAlgo TKBRep TKGeomBase) 
target_link_libraries(sphinxsys_opencascade PUBLIC TKG3d TKG2d TKMath TKernel) 
```
If you leave all opencascade headers in a catch-all headers "opencascade.h" then all those targets must be in the public interface, which creates more coupling. It has relatively little effect on a dev environment, but if sphinxsys is hoping to be packaged at some point, that will prevent some headache.

FabienPean-Virtonomy: Also could you please resolve conversation of comments you integrated to help reviewer keeping track of the PR status

yuanjiajy: ok I will change that based on your original one.

Xiangyu-Hu (approved):

PR #382, opened 2023-07-26
fix bugs, may be still other bugs, please test it.

Xiangyu-Hu:
fix bugs, may be still other bugs, please test it.


Shuaihao-Zhang:
> fix bugs, may be still other bugs, please test it.

The result seems ok. I have added a regression test for this case.

PR #373, opened 2023-07-18
[SYCL] Fluid density relaxation

nR3D:
- Implemented `BaseIntegration2ndHalfWithWalll` kernel class to enable device execution of density relaxation

Here is the benchmark results with 80'000 particles:

```
SYCL fluid_density_relaxation computation: 0.27803 seconds
Reference fluid_density_relaxation computation: 13.0099 seconds
Speedup fluid_density_relaxation (main loop): 46.793x
```

PR #397, opened 2023-08-01
[SYCL] Writing to files

nR3D:
Classes `BodyStatesRecordingToVtp`, `TotalMechanicalEnergy` and `ObservedQuantityRecording` have been adapted to work with data computed on a SYCL device. In general, output files are saved on host storage, which means that variables residing on a GPU device need to be first moved to host memory in order to be then saved to file.
- In the case of `BodyStatesRecordingToVtp`, and specifically considering the test case Dambreak, variables stored inside `BaseParticles::variables_to_write_` are now first copied from device to host, and then printed to file. It's worth noting that some other test cases may need to transfer and compute `BaseParticles::derived_variables_` as well.
- `TotalMechanicalEnergy` is computed using a reduction, hence the only requirement to enable its execution with SYCL was to add the corresponding kernel class `TotalMechanicalEnergyKernel`. Another approach would have been to first copy the corresponding variables (mass, velocity, and position) to host and then compute the reduction but this would mean many data transfers at every file writing (which happen inside the main loop).
- Regarding `ObservedQuantityRecording`, one of its base classes, `InteractionDynamics`, was previously executed with the default `ParallelPolicy`, i.e. TBB parallel execution. A new `ExecutionPolicy` template parameter for `ObservedQuantityRecording` and `ObservedAQuantity` has been added to explicitly define such policy. Kernel class for class `BaseInteraction` has also been added in order to run `ObservedQuantityRecording` with a parallel SYCL policy.

Other changes:
- `DataTypeIndex` bug mentioned by @ChiZhangatTUM in #373 has been solved by relying on `std::conditional` instead of `std::enable_if`
- Helper class `DataTypeEquivalence` has been added to represent type equivalence between host and device (e.g. a `Real` on host corresponds to a `DeviceReal` on device). This helps in those cases in which the host type has been passed as a template typename and the equivalent device type is also needed; for instance, such case happens for `BaseInteraction`.
- Memory transfer of particle configurations have been moved out of `ComplexRelation` constructor, since allocation will probably be moved inside `Inner/ContactRelation` constructor, and copying will be unnecessary once cell linked list will be computed inside the device memory.
- SYCL `particle_for` is using an user defined work group size, but such value needs to be larger than the number of particles computed, hence a minimum size has been defined. In addition, the iteration size need to be a multiple of the work group size, but this check has not yet been implemented, so it's still up to the user to pass the right work group size according to the number of particles computed.

### DIFF
--- a/src/shared/body_relations/complex_body_relation.cpp
+++ b/src/shared/body_relations/complex_body_relation.cpp
@@ -28,12 +28,6 @@ ComplexRelation::ComplexRelation(RealBody &real_body, RealBodyVector contact_bod
       contact_configuration_(contact_relation_.contact_configuration_)
 {
     resizeConfiguration();
-
-    inner_relation_.allocateInnerConfigurationDevice();
-    inner_relation_.copyInnerConfigurationToDevice();
-
-    contact_relation_.allocateContactConfiguration();
-    contact_relation_.copyContactConfigurationToDevice();
 }
 //=================================================================================================//
 ComplexRelation::
--- a/src/shared/common/base_data_type.h
+++ b/src/shared/common/base_data_type.h
@@ -119,7 +119,7 @@ struct ZeroData<int>
     static inline int value = 0;
 };
 /** Type trait for data type index. */
-template <typename T, typename Enable = void>
+template <typename T, class Enable = std::true_type>
 struct DataTypeIndex
 {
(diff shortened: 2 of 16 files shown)


Xiangyu-Hu:
@nR3D Well done, I am looking forward to the great achievement of the project!

ChiZhangatTUM:
@nR3D Very impressive work.

PR #398, opened 2023-08-01
Without global variable

Xiangyu-Hu:
1) revise the transport velocity formulation so  that it correct all particles for multiphase flow. 
2) revised the droplet and wetting case so that the simulation can be stable to long physical time.

### DIFF
--- a/src/shared/particle_dynamics/base_local_dynamics.h
+++ b/src/shared/particle_dynamics/base_local_dynamics.h
@@ -36,33 +36,76 @@
 
 namespace SPH
 {
-/** A Functor for Summation */
+//----------------------------------------------------------------------
+// Particle group scope functors
+//----------------------------------------------------------------------
+class AllParticles
+{
+  public:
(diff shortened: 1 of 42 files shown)


Shuaihao-Zhang:
The result seems good for the case `test_2d_square_droplet.`

For the `test_2d_wetting_effects` case, the particle distribution is not uniform. Maybe other reasons related to the boundary condition.
image

I found two other problems, which may be not related to the current modification.
(1) The first one is that, for the `test_2d_wetting_effects` case, the air phase and the water phase share the same layer of particle on the interface, as shown below:
image
I found this problem when increasing the resolution. For very low resolutions, we don't have this problem. I guess the bug is from the particle generation process.

(2) The second one is that the advection time step is always the same as the acoustic time step for the two cases.
image

Xiangyu-Hu:
@Shuaihao-Zhang probably the transport velocity correction is too much.

Shuaihao-Zhang (approved):

PR #401, opened 2023-08-05
Xiangyu/staggered surface tension

Xiangyu-Hu:
a special surface tension model for checking.
@Shuaihao-Zhang


Shuaihao-Zhang:
This method is quite good, the result is stable even when time t=100. Is there any reference about this method?

One thing is that, it seems when with high resolutions, the air particles and water particles will mix with each other near the interface.  The figure shows the result with a high resolution at time t=10.
image

Shuaihao-Zhang:
> We still need to revise it for better this issue.
> 
Sure, I will firstly check the current version in detail.

PR #409, opened 2023-08-17
add the VTP file in FVM method

ZhentongWang:
add the VTP file in FVM method and change the output files in the sources. The change is removing the output of particle's position from base_particles.hpp to io_vtk.cpp file due to the fact that the definition of position of particles and cells are different. After removing, the output of other variables are the same.

Besides, there is small modification in FVM part about max and min distance between nodes.


Xiangyu-Hu (commented):
Does the change influence the output files of other cases?

ZhentongWang:
> Does the change influence the output files of other cases?

No, this doesn't. Other cases still  enable output original results successfully.

Xiangyu-Hu (approved):

PR #414, opened 2023-08-24
Introduce tinyxml2 parser, other than use xml engine from simbody. 

ChiZhangatTUM:
XML parser based on tinyxml2.
Done: Particle restart & reload are implemented. 
TBD: regression test is not implemented.


ChiZhangatTUM:
> @ChiZhangatTUM There is a case did not pass, probably due to the change of body name?

Maybe due to the particle relax and reload parameter setup, I will go to detail this evening.

ChiZhangatTUM:
> @ChiZhangatTUM There is a case did not pass, probably due to the change of body name?

Seems that particle reload is fine. The simulation stop due to minus time step size. I will have more checks.

ChiZhangatTUM:
> @ChiZhangatTUM There is a case did not pass, probably due to the change of body name?
Seems that particle reload is fine. The simulation stop due to minus time step size. I will have more checks.

ChiZhangatTUM:
> @ChiZhangatTUM There is a case did not pass, probably due to the change of body name?

Seems that particle volume is write as zero, I wilk found out the bug.

src/shared/tools/tinyxml2.cpp
Xiangyu-Hu, line 1: Should this included in VCPKG, not a embedded in source files?

src/shared/tools/xml_parser.h
```diff
@@ -0,0 +1,387 @@
...
+	// Eigen::IOFormat HeavyFmt(Eigen::FullPrecision, 0, ", ", ";\n", "[", "]", "[", "]");
+
+	template < typename DataType >
+	inline std::string DataToString( const DataType &value )
+	{
+		std::ostringstream out;
+    	out.precision(15);
+    	out << std::fixed << value;
+    	return std::move(out).str();
+		//return std::to_string(value);
```
ChiZhangatTUM, line 63: Another option : 

template<typename T> 
std::string data_to_string(const T& value)
{
    std::array<char, 15> str;

    auto res = std::to_chars(str.data(), str.data() + str.size(), value);
    return std::string(str.data(), res.ptr);
}

Seems not work due to the compiler version.

src/shared/tools/tinyxml2.cpp
ChiZhangatTUM: Use vcpkg installation is one option, but I thought that use " embedded source file " is more flexible as stated by tinyxml2 project that "TinyXML-2 is designed to be easy and fast to learn. It is one header and one cpp file. Simply add these to your project and off you go. "
Anyway, if you prefer vcpkg installation, it can be changed with minimal effort.

Xiangyu-Hu:
> XML parser based on tinyxml2. Done: Particle restart & reload are implemented. TBD: regression test is not implemented.

I every thing finished now?

ChiZhangatTUM:
> > XML parser based on tinyxml2. Done: Particle restart & reload are implemented. TBD: regression test is not implemented.
> 
> I every thing finished now?

In this PR, only the particle writer and reloader is rewritten with the new xml parser to make sure that all work properly, then in the next PR the regression test will be modified with all test data are re-generated.

Xiangyu-Hu:
> > > XML parser based on tinyxml2. Done: Particle restart & reload are implemented. TBD: regression test is not implemented.
> > 
> > 
> > I every thing finished now?
> 
> In this PR, only the particle writer and reloader is rewritten with the new xml parser to make sure that all work properly, then in the next PR the regression test will be modified with all test data are re-generated.

How this unfinished things influence the CI tests if one may generate new data set?

ChiZhangatTUM:
> > > > XML parser based on tinyxml2. Done: Particle restart & reload are implemented. TBD: regression test is not implemented.
> > > 
> > > 
> > > I every thing finished now?
> > 
> > 
> > In this PR, only the particle writer and reloader is rewritten with the new xml parser to make sure that all work properly, then in the next PR the regression test will be modified with all test data are re-generated.
> 
> How this unfinished things influence the CI tests if one may generate new data set?

No effect, as the current regression  test is using the old xml engine, which is independent on the new xml2 parser.

Xiangyu-Hu (approved):

PR #429, opened 2023-09-06
add viscous solid model and corresponding case

Xiaojingtang1234:
the master thesis case
Add viscous solid model and corresponding case
this is the result
image
image


Shuaihao-Zhang (approved):

Shuaihao-Zhang:
It seems that there is the hourglass issue. Maybe later can try to solve that.

Shuaihao-Zhang:
image
I checked the stress distribution. Hourglass modes are obvious and the tensile instability also occurs because of this.

PR #430, opened 2023-09-06
switch command option in cases to default

Xiangyu-Hu:
@SPHYaru I remember, you have mentioned a bug on vorticity calculation. Could you specify it? Thanks.


Lisu-lisa:
Ok

Lisu-lisa (approved):

SPHYaru (approved):

SPHYaru:
> @SPHYaru I remember, you have mentioned a bug on vorticity calculation. Could you specify it? Thanks.

It appears to be a small bug. When calculating the vorticity, we should use "InteractionDynamics<fluid_dynamics::VorticityInner> compute_vorticity(water_block_complex.getInnerRelation());" instead of "InteractionDynamics<fluid_dynamics::VorticityInner> compute_vorticity(water_block_inner);" It was my mistake.

PR #434, opened 2023-09-16
Thermal diffusion refactoring. 

ChiZhangatTUM:
1, introduce thermal diffusion material, whose constant parameter should be density, capacity and conductivity, and thermal diffusivity is the thermal conductivity divided by density and specific heat capacity. 
2, thermal diffusion dynamics is add, as its accurate SPH approximation is  rho * cp * (dT/dt) = 2 * sum (2.0 k1 * k2 / (k1 + k2)) * grad w * V, which consider the variation of density and exhibits better performance when different conductivity involved. 
3, contact dynamics is added for different conductivity.

### DIFF
--- a/src/shared/materials/contact_diffusivity.h
+++ b/src/shared/materials/contact_diffusivity.h
@@ -1,0 +1,102 @@
+/* ------------------------------------------------------------------------- *
+ *                                SPHinXsys                                  *
+ * ------------------------------------------------------------------------- *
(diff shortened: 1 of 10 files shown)


Xiangyu-Hu:
@ChiZhangatTUM I think that we should have a discussion on this.

Xiangyu-Hu:
@ChiZhangatTUM as I am thinking using template design on diffusion materials.

PR #448, opened 2023-09-29
[SYCL] Cell linked list and system configuration support

nR3D:
# Main changes
Cell linked list updates are now computed and stored on device. Thanks to this, neighboring information may be computed on device too, hence inner and contact configurations are updated without copying data to host.

## Details

Cell linked list on host is stored as a concurrent vector of a list of indices. The concurrency is used to update the list (i.e. cell) while iterating in parallel through particles.
Device kernel is structured in a similar way, but since SYCL does not provide a concurrent vector, and access to memory is left to be managed manually, atomics are used instead. Another difference is the impossibility to allocated device memory during kernel execution, as it would required a vector of lists. For this reason two arrays are used to represent the cell linked list: a _head index list_ and an _index list_. The former is an array that stores one index per cell, its values represent the index of the first particle of each cell, such value is the first index to be iterated in _index list_; the latter is effectively a forward linked list, which stores N values, where N = `total_real_particles_`, and each value is the index of the next value to be dereferenced. 

> **Note:** End of the list, in both cases, is represented by zero; for this reason, all indices stored in both lists are increased by 1, so that the value 0 is reserved for the end-of-list case.

Since the order in which particles are stored inside a cell is not relevant, the fastest atomic memory order `relaxed` is used.

## Benchmark

Interval times of dambreak run with 80'000 particles:

Reference TBB (CPU, Intel Xeon E5):
```
Total wall time for computation: 1148.758320495 seconds.
interval_computing_time_step =34.634967447
interval_computing_fluid_pressure_relaxation = 609.707733325
interval_updating_configuration = 380.527507446
```

SYCL (GPU, NVIDIA RTX 2080 Ti):
```
Total wall time for computation: 364.482508205 seconds.
interval_computing_time_step =8.011819663
interval_computing_fluid_pressure_relaxation = 95.911848134
interval_updating_configuration = 140.936792402
```

# Other changes

- Dambreak test file has been restored to its `master` branch version, while the SYCL version of dambreak has been moved to a new test-case: `tests/2d_examples/test_2d_dambreak_sycl`. This will help comparing both execution time and results.
- To reduce device memory allocations when execution is carried out only on host, kernel classes are initialized with device allocated memory only when the system configuration is device allocated. Otherwise `nullptr` will be passed to kernel classes, since they will never be executed.
    - Note: a better way to handle this case is to construct kernel classes only when the class will be executed on device, but parameters to determine such choice are not always available while constructing the kernel class. A solution would be to construct the kernel class the first time it needs to be executed on device, in a lazy manner.
    - Thanks to this change now the SYCL benchmark test-case can run twice the number of particles.
- Added a method to explicitly copy device memory before writing to file. The copy could be moved inside methods `body_states_recording.writeToFile()`, but having a separate method could be useful if the copy operation is executed in parallel with some other process. Although, this would need some changes to the `copyDataFromDevice` interface, to return the SYCL event to wait on.


ChiZhangatTUM:
@nR3D Well done.

src/for_2D_build/common/data_type.h
```diff
@@ -36,6 +36,7 @@
 namespace SPH
 {
 using Arrayi = Array2i;
+using DeviceArrayi = DeviceArray2i;
```
ChiZhangatTUM, line 39: As you defined DeviceArrayi here, does that means all other variables without "Device" term are HostVaribles?

nR3D: Yes exactly, device variables are using some SYCL built-in datatypes that are optimized for kernel executions. For example DeviceArray2i is `sycl::vec<int,2>`, which will be translated to CUDA vectors at compile-time (for NVIDIA backends for example). I still have not tried it, but it should be possible to set device variables = host variables, it's just that some functions calls would need to be changed (for example Eigen vectors use dot products as `v1.dot(v2)`, while SYCL interface is `sycl::dot(v1, v2)`). 

I already have a branch in which I am trying to set some helper functions that select the right function call depending on the vector's type, something like:
```
auto dot(Eigen::vec v1, Eigen::vec v2) { return v1.dot(v2); }
auto dot(sycl::vec v1, sycl::vec v2) { return sycl::dot(v1, v2); } 
```

So that, within the code, you can just call `dot(v1, v2)` and it would select the right interface.

ChiZhangatTUM: Great! A helper is definitive essential. Looking forward to seeing it in you next implementation.

src/shared/particle_neighborhood/neighborhood.h
```diff
@@ -74,7 +75,7 @@ using ParticleConfiguration = StdLargeVec<Neighborhood>;
 class NeighborhoodDevice {
   public:
     size_t* current_size_;	/**< the current number of neighbors */
-    static constexpr size_t allocated_size_ = 40;  /**< the limit of neighbors does not require memory allocation  */
+    static constexpr size_t allocated_size_ = 28;  /**< the limit of neighbors does not require memory allocation  */
```
ChiZhangatTUM, line 78: Why do this change?

src/shared/meshes/cell_linked_list.h
```diff
@@ -112,6 +155,16 @@ class CellLinkedList : public BaseCellLinkedList, public Mesh
...
+    template<class ExecutionPolicy>
+    StdLargeVec<size_t> &computingSequence(BaseParticles &base_particles, ExecutionPolicy)
+    {
+        /*StdLargeVec<Vecd> &pos = base_particles.pos_;
+        StdLargeVec<size_t> &sequence = base_particles.sequence_;
+        size_t total_real_particles = base_particles.total_real_particles_;
+        particle_for(execution_policy, total_real_particles, [&](size_t i, auto&& kernel)
+            { sequence[i] = transferMeshIndexToMortonOrder(CellIndexFromPosition(pos[i])); }, ExecutionSelector<BaseMesh, ExecutionPolicy>(this));
+        return sequence;*/
+    }
```
ChiZhangatTUM, line 167: Seems that all the contents in this functions is commented.

src/shared/common/base_data_type.h
```diff
@@ -122,6 +124,22 @@ struct DataTypeEquivalence<TypeVec3d, enable_if_is_either_t<TypeVec3d, Vec3d, De
...
+    using host_t = Array2i;
+    using device_t = DeviceArray2i;
+};
+
+template<class TypeArray3i>
+struct DataTypeEquivalence<TypeArray3i, enable_if_is_either_t<TypeArray3i, Array3i , DeviceArray3i>> {
+    using host_t = Array3i;
+    using device_t = DeviceArray3i;
+};
+
```
ChiZhangatTUM, line 138: Variable name host_t and device_t maybe changed to host_type and device_type for clarity.

src/shared/body_relations/base_body_relation.h
```diff
@@ -163,22 +169,27 @@ class BaseContactRelation : public SPHRelation
...
-                    contact_configuration_.at(k).at(i) = contact_configuration_device_.at(k).at(i);
-        }
-	};
+    bool isDeviceConfigurationAllocated() { return device_configuration_allocated_; }
+
+    void allocateContactConfiguration() {
+        for (const auto & body : contact_configuration_)
+            contact_configuration_device_.emplace_back(body.size(), execution::executionQueue.getQueue());
+        device_configuration_allocated_ = true;
+    }
```
ChiZhangatTUM, line 178: allocatedConrtactConfigurationOnDevice for DeviceallocatedConrtactConfiguration?

ChiZhangatTUM: > allocatedConrtactConfigurationOnDevice for DeviceallocatedConrtactConfiguration?
allocatedConrtactConfigurationOnDevice or DeviceallocatedConrtactConfiguration?

src/shared/particle_neighborhood/neighborhood.h
nR3D: New memory cannot be allocated within kernel execution, for this reason an upper bound of neighbors needs to be defined. I tested that, in the 2d case, neighbors are around 26, hence 40 was giving too much slack and wasted device memory. 28 is a better upper bound.

To be fair, those values don't need to be constexpr, they can also be computed at runtime when the neighborhood is first defined.

Since the maximum number of neighbors depends on the smoothing length, a better approach would be to implement a function returning the upper bound based on the currently used smoothing kernel. But I have not checked how coupled the smoothing kernel and the neighborhood construction are.

src/shared/meshes/cell_linked_list.h
nR3D: Yes, I should have kept it for the next PR instead of commenting it.
Essentially, I am now working on particle sorting on sycl, and since sorting is also connected to the cell linked list I started working on both at the same time, so some code regarding the next PR slipped into this one.

src/shared/common/base_data_type.h
nR3D: C++ standard usually appends `_t` when referring to a type (`size_t` for instance). So I thought it would closer to the C++ jargon.

But if `_type` is clearer I can change it.

src/shared/body_relations/base_body_relation.h
nR3D: I would say `allocateContactConfigurationOnDevice`, if it's not too lengthy

src/shared/common/base_data_type.h
ChiZhangatTUM: That's ok, you can change it next PR.

src/shared/body_relations/base_body_relation.h
nR3D: Are you referring to the method's name or the name of variable `device_configuration_allocated_`?

ChiZhangatTUM (approved):
Well done.

Issue #449, opened 2023-10-06
Shell material issue strain/deformation gradient


FabienPean-Virtonomy:
The arguments (almansi strain and deformation gradient) passed to the constitutive laws classes are not directly related together.

The discrepancy originates in the integration 1st half for shells:

https://github.com/Xiangyu-Hu/SPHinXsys/blob/abae343aface52952af98ed95a7915bb8e62f1fc/src/shared/particle_dynamics/solid_dynamics/thin_structure_dynamics.cpp#L135-L147

To reproduce a failure, it suffices to replace B in
https://github.com/Xiangyu-Hu/SPHinXsys/blob/abae343aface52952af98ed95a7915bb8e62f1fc/src/shared/materials/elastic_solid.cpp#L93-L101

by 
```diff
-     Matd B = (-2.0 * almansi_strain + Matd::Identity()).inverse(); 
+    const SPH::Mat3d B = F * F.transpose();
```
would generate a failed simulation.

Ideally, the API should allow to work like this, so that user can implement only StressPK2 or StressCauchy:
```cpp
void StressCauchy(const Mat3d& F, size_t index_i)
{
    double J = F.determinant();
    return F*StressPK2(F)*F.transpose()/J;
}
```

PR #450, opened 2023-10-08
Correct shell material issue.

DongWuTUM:
It would cause a problem when shell model use other material in the future as shell model uses the corrected Almansi strain.


DongWuTUM:
This pull request Closes https://github.com/Xiangyu-Hu/SPHinXsys/issues/449

Xiangyu-Hu (approved):

PR #451, opened 2023-10-08
clarify that base material only referred from base particle other than sph body

Xiangyu-Hu:


Xiangyu-Hu:
I later think that this is not a good idea as it is against the principle of loss coupling with the library.

PR #454, opened 2023-10-14
[SYCL] Particle sorting and homogenized ND range kernels

nR3D:
# Main changes

- Particle sorting is now offloaded to GPU, making it possible to run the main simulation steps entirely on SYCL.
    - The sorting algorithm is based on an implementation of [radix sort optimized for accelerators](https://doi.org/10.1109/IPDPS.2009.5161005).
- Kernel executions are now homogenized w.r.t. the set work-group size.
    - This gives more freedom in the choice of the work-group size, which was before required to be a divisor of the number of particles computed, while now the choice can be arbitrary.

## Benchmark

Benchmark of 80'000 particles:
```
Total wall time for computation: 314.42814360 seconds.
interval_computing_time_step =8.077749454
interval_computing_fluid_pressure_relaxation = 96.600376022
interval_updating_configuration = 89.765419521
```

Compared to [previous results](https://github.com/Xiangyu-Hu/SPHinXsys/pull/448) this marks a 1.15x speed-up, and 3.65x speed-up w.r.t. TBB.

Tested with 500'000 particles, results are:

Reference:
```
Total wall time for computation: 17098.949532568 seconds.
interval_computing_time_step =514.735826599
interval_computing_fluid_pressure_relaxation = 9491.617997688
interval_updating_configuration = 5973.237301082
```

SYCL:
```
Total wall time for computation: 3616.444860880 seconds.
interval_computing_time_step =62.580767842
interval_computing_fluid_pressure_relaxation = 1040.480258746
interval_updating_configuration = 1375.825838244
```

Hence, 4.73x speed-up w.r.t. multi-core TBB execution.

### DIFF
--- a/src/shared/common/base_data_type.h
+++ b/src/shared/common/base_data_type.h
@@ -100,7 +100,7 @@ using DeviceArray3i = sycl::int3;
 
 template<typename Type, class Enable = void>
 struct DataTypeEquivalence {
-    static constexpr bool exists = false;
+    static constexpr bool type_defined = false;
     static_assert("Type non recognized as host or device type.");
 };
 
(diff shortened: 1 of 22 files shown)


ChiZhangatTUM:
Well done.

src/shared/common/base_data_type.h
ChiZhangatTUM, line 1: Can you specify the meaning of  variable "exists" added in the following code.

ChiZhangatTUM:
Great work. A suggestion would be introducing a helper of SYCL_MACRO other than using it directly in the SPHinXsys code. Please think about this in your following work.

src/shared/common/base_data_type.h
nR3D: It specifies if a data type equivalence between host and device for a given type exists.
In a method like `BaseParticles::registerSortableVariable` it has been used to determine if, aside from the normal host variable, there is also an equivalent device type that can be registered, since host and device variables are stored in different variables.
An alternative name for it would be `defined` or `is_defined`.

ChiZhangatTUM: Maybe " is_registered" ?

nR3D: The idea is to determine at compile-time whether an equivalence between an host type and a device type is defined.
This concept is then used to register sortable variables, but it is not strictly related to that.
For instance, Real has a corresponding type DeviceReal, so `DataTypeEquivalence<Real>::exists == true`, and  `DataTypeEquivalence<Real>::device_t == DeviceReal`, while there is not an equivalent device type for Rotation2d, hence `DataTypeEquivalence<Rotation2d>::exists == false`, otherwise you would get a compile-time error trying to access `DataTypeEquivalence<Rotation2d>::device_t`.

ChiZhangatTUM: Understood. How about type_exsited ?

nR3D: I feel like I should have used "defined" from the start instead of "extsts", given that I am looking for a type definition.

Maybe `type_defined`?

Xiangyu-Hu:
> Great work. A suggestion would be introducing a helper of SYCL_MACRO other than using it directly in the SPHinXsys code. Please think about this in your following work.

I would prefer using the strategy of execution policy. On the other hand, we need migrate the entire SPHinXysy to  be computational-kernel based. So there is not difference on the code no matter it is running on CPU or GPU.

src/shared/common/base_data_type.h
ChiZhangatTUM: Seems an good option.

ChiZhangatTUM (approved):

PR #455, opened 2023-10-18
tested dambreak and elastic gate

Xiangyu-Hu:
The new limiter is symmetric to compression and expansion.


Bo-Zhang1995:
I have tested the dambreak 2d with 1/4 and 1/8 current resolution, and they can run and the result don't show strange things.
This new scheme may works.

Xiangyu-Hu:
I think that the original formulation is also continuous and seems fit better to the physics of Riemann problem. Also may be the non-symmetric for biased to compression is physically more sound. I will consider it latter when issues prompted out.

PR #456, opened 2023-10-19
[SYCL] Event-driven kernel execution

nR3D:
# Changes

- A new class, `ExecutionEvent`, offers the possibility to launch multiple GPU and CPU tasks asynchronously by defining the dependencies that exist between them.

# Details

So far, computations have been offloaded to SYCL device by submitting a kernel to a queue and waiting for it to finish.
However, some methods are executed at a given moment but their result will not be relevant until much later. This is the case for system configurations update and writing output files, in the latter case their results are never required by the following iterations.
SYCL provides an event class that can be used to postpone kernel synchronization: it is possible to submit a kernel, then continue with other computations not dependent on such execution, and finally wait for the event to complete when the result is actually necessary.
Additionally, SYCL offers a command group directive, `host_task`, that executes a function on a new CPU thread. The function in this case is not tied to the same restrictions applied to the kernels submitted to the device, instead any C++ code can be run as an host task.
`ExecutionEvent` combines those two features, offering methods to build a list of dependencies coming from SYCL kernels, and the possibility to set an host task to be run asynchronously when all the device events have been completed.

An example would be `body_state_recording`, that first needs to copy data from device to host, and then it can write output files with CPU:
```c++
ExecutionEvent async_body_states_copy_event = body_states_recording.copyDeviceData();
// ... other computations hiding the transfer runtime
async_body_states_copy_event.then([&] { body_states_recording.writeToFile(); }).wait();
```

It is worth noting that the task expressed by the lambda function passed to `.then()` will be executed on a different thread than the rest of the code, which means that, while the files are being processed, the rest  of the simulations carries on with the remaining iterations. Effectively, the main CPU thread will be responsible for submitting kernels to the device, while the threads managed by SYCL will execute the major computational effort that, so far, slowed-down computations.

# Benchmarks

> Refer to #454 and #448 for reference runtime on CPU

_SYCL with 500'000 particles_ 
```
Total wall time for computation: 2964.484684851 seconds.
interval_computing_time_step =1410.974875226
interval_computing_fluid_pressure_relaxation = 1048.459943614
interval_updating_configuration = 260.050099024
interval_writing_files = 228.746797083
```

_SYCL with 80'000 particles_
```
Total wall time for computation: 258.931707035 seconds.
interval_computing_time_step =83.856230888
interval_computing_fluid_pressure_relaxation = 99.339878925
interval_updating_configuration = 40.913803999
interval_writing_files = 30.766704951
```

Which corresponds to 5.7x and 4.4x speed-up respectively compared to multi-threaded TBB

> *Note:* `interval_updating_configuration` and `interval_writing_files` show a shorter time interval because they now share computational resources with other steps, such as time-step computations, that is now running simultaneously with configuration update, hence slowing down time-step calculation, but improving the overall performance.

### DIFF
--- a/src/for_2D_build/meshes/cell_linked_list.hpp
+++ b/src/for_2D_build/meshes/cell_linked_list.hpp
@@ -45,7 +45,7 @@ void CellLinkedList::searchNeighborsByParticles(
 }
 //=================================================================================================//
 template <class DynamicsRange, typename GetSearchDepth, typename GetNeighborRelation>
-ExecutionEvent CellLinkedListKernel::searchNeighborsByParticles(
+execution::ExecutionEvent CellLinkedListKernel::searchNeighborsByParticles(
     DynamicsRange &dynamics_range, NeighborhoodDevice *particle_configuration,
     GetSearchDepth &get_search_depth, GetNeighborRelation& get_neighbor_relation,
     execution::ExecutionEvent dependency_event)
--- a/src/for_2D_build/meshes/cell_linked_list.hpp
+++ b/src/for_2D_build/meshes/cell_linked_list.hpp
@@ -45,19 +45,21 @@ void CellLinkedList::searchNeighborsByParticles(
 }
 //=================================================================================================//
 template <class DynamicsRange, typename GetSearchDepth, typename GetNeighborRelation>
-void CellLinkedListKernel::searchNeighborsByParticles(
+ExecutionEvent CellLinkedListKernel::searchNeighborsByParticles(
     DynamicsRange &dynamics_range, NeighborhoodDevice *particle_configuration,
-    GetSearchDepth &get_search_depth, GetNeighborRelation& get_neighbor_relation)
+    GetSearchDepth &get_search_depth, GetNeighborRelation& get_neighbor_relation,
+    execution::ExecutionEvent dependency_event)
 {
     auto *pos = dynamics_range.getBaseParticles().template getDeviceVariableByName<DeviceVec2d>("Position");
     auto get_neighbor_relation_kernel = *get_neighbor_relation.getDeviceProxy().getKernel();
     const int search_depth = get_search_depth(0);
     const auto loop_range = dynamics_range.LoopRange();
-    executionQueue.getQueue()
+    return executionQueue.getQueue()
         .submit(
             [&, index_list=index_list_, index_head_list=index_head_list_, list_data_pos=list_data_pos_,
              list_data_Vol=list_data_Vol_, mesh_lower_bound=mesh_lower_bound_,
              grid_spacing=grid_spacing_, all_grid_points=all_grid_points_, all_cells=all_cells_](sycl::handler &cgh) {
+                cgh.depends_on(dependency_event.getEventList());
                 cgh.parallel_for(executionQueue.getUniformNdRange(loop_range), [=](sycl::nd_item<1> it) {
                                      const size_t index_i = it.get_global_id(0);
                                      if(index_i < loop_range)
@@ -81,7 +83,7 @@ void CellLinkedListKernel::searchNeighborsByParticles(
                                              });
                                      }
                                  });
-            }).wait_and_throw();
+            });
 }
 //=================================================================================================//
 } // namespace SPH
--- a/src/shared/bodies/base_body.h
+++ b/src/shared/bodies/base_body.h
@@ -243,19 +243,19 @@ class RealBody : public SPHBody
     bool getUseSplitCellLists() { return use_split_cell_lists_; };
     SplitCellLists &getSplitCellLists() { return split_cell_lists_; };
     template<class ExecutionPolicy = execution::ParallelPolicy>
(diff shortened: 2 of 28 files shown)


src/for_2D_build/meshes/cell_linked_list.hpp
```diff
@@ -45,19 +45,21 @@ void CellLinkedList::searchNeighborsByParticles(
 }
 //=================================================================================================//
 template <class DynamicsRange, typename GetSearchDepth, typename GetNeighborRelation>
-void CellLinkedListKernel::searchNeighborsByParticles(
+ExecutionEvent CellLinkedListKernel::searchNeighborsByParticles(
```
ChiZhangatTUM, line 48: No namespace of "execution" here?

nR3D: You are right, it should have it, I'll fix it. Curious how it doesn't throw errors or warnings at compile time. Probably because particle_iterators.h is included in cell_linked_list.hpp, which has the using namespace directive for execution.

ChiZhangatTUM (approved):

src/for_2D_build/meshes/cell_linked_list.hpp
ChiZhangatTUM: thx, I approved this PR.

nR3D:
I just noticed that I missed a wait() that becomes relevant for a higher number of particles, I'll fix it later, for the moment I set this PR as draft.

ChiZhangatTUM:
> I just noticed that I missed a wait() that becomes relevant for a higher number of particles, I'll fix it later, for the moment I set this PR as draft.

Ok

nR3D:
Issue solved.
Initially I remembered that some I/O classes had memory copies from device to which I forgot to add a wait() , but once I did I started to notice some concurrency problems. It was due to different methods trying to copy to the same variable at the same time, so I added the possibility to retrieve the sycl::event returned by host_task to be sure that two methods writing to the same variables were not executed at the same time. Moreover, I reduced the number of variables copied by BodyStatesRecording::copyDeviceData in order to transfer only those variables that will be actually written and not all the particle variables.
By doing so I also realized that I never wrote a methods to copy restart variables from the device, since I always had an entire copy from device every loop, and now that it's no longer the case I implemented a method to copy only restart variables from device. Since RestartIO and BodyStatesRecording might be copying the same variable, I added an event dependency between the two. This slows down slightly the final runtime, but the results are still better than before. I will now update the above benchmarks.

ChiZhangatTUM:
> Issue solved. Initially I remembered that some I/O classes had memory copies from device to which I forgot to add a wait() , but once I did I started to notice some concurrency problems. It was due to different methods trying to copy to the same variable at the same time, so I added the possibility to retrieve the sycl::event returned by host_task to be sure that two methods writing to the same variables were not executed at the same time. Moreover, I reduced the number of variables copied by BodyStatesRecording::copyDeviceData in order to transfer only those variables that will be actually written and not all the particle variables. By doing so I also realized that I never wrote a methods to copy restart variables from the device, since I always had an entire copy from device every loop, and now that it's no longer the case I implemented a method to copy only restart variables from device. Since RestartIO and BodyStatesRecording might be copying the same variable, I added an event dependency between the two. This slows down slightly the final runtime, but the results are still better than before. I will now update the above benchmarks (edit: benchmarks updated).

Thx. If I understood your info correctly. Now, only specified variables are copied from device and written to files. The users may not specify one variable to be written in the output file with simple modification in the main file, as CPU code of add variable to the write list?

nR3D:
> @ChiZhangatTUM
> Thx. If I understood your info correctly. Now, only specified variables are copied from device and written to files. The users may not specify one variable to be written in the output file with simple modification in the main file, as CPU code of add variable to the write list?

You are right, what I did was to manually copy those variables that are always written to file, such as position and velocity, but a better approach is to use `BaseParticles::variables_to_write_`. I have just committed a small change to iterate over `variables_to_write_` and copy those that are also registered as device variables.
Users may  still add variables to write using `BaseParticles::addVariableToWrite`, there is no distinction between host variables and device variables in this case.
Similarly, RestartIO uses `variables_to_restart_` to copy only the necessary variables.
This PR has already been merged, so I will add the changes to the next one.

ChiZhangatTUM:
> > @ChiZhangatTUM
> > Thx. If I understood your info correctly. Now, only specified variables are copied from device and written to files. The users may not specify one variable to be written in the output file with simple modification in the main file, as CPU code of add variable to the write list?
> 
> You are right, what I did was to manually copy those variables that are always written to file, such as position and velocity, but a better approach is to use `BaseParticles::variables_to_write_`. I have just committed a small change to iterate over `variables_to_write_` and copy those that are also registered as device variables. Users may still add variables to write using `BaseParticles::addVariableToWrite`, there is no distinction between host variables and device variables in this case. Similarly, RestartIO uses `variables_to_restart_` to copy only the necessary variables. This PR has already been merged, so I will add the changes to the next one.

That would be great. thx.

PR #462, opened 2023-10-30
[SYCL] Enable double precision and generalized vector operations

nR3D:
# Changes
- `DeviceReal` can now be set to `double`
    - This change required just some type casting to make sure that the right types could be inferred at compile-time.
- As discussed #448, some helper functions regarding vectors have been added to generalize kernel executions
    - > I have not included a few more changes that would be needed to let `DeviceVecd = Vecd` (i.e. using Eigen matrices as device vectors), since Eigen classes are not device-copyable, hence they cannot be run within kernels without other significant changes to how vectors are stored. I have added the changes needed to enable `DeviceVecd = Vecd` to https://github.com/nR3D/SPHinXsys/tree/sycl-eigen in case they would ever be needed. They mainly involve the definition of other helper classes for vectors, and changes to the declaration of some templates.

# Benchmarks

With 300'000 particles, using double-precision makes the simulation 86.4% slower than with single-precision

Double:
```
Total wall time for computation: 1114.259535746 seconds.
interval_computing_time_step =768.747579194
interval_computing_fluid_pressure_relaxation = 325.318849561
interval_updating_configuration = 3.267941609
interval_writing_files = 10.581363697
```

Float:
```
Total wall time for computation: 597.805587192 seconds.
interval_computing_time_step =319.801477432
interval_computing_fluid_pressure_relaxation = 259.038648931
interval_updating_configuration = 2.944420347
interval_writing_files = 10.566627757
```

### DIFF
--- a/src/shared/common/vector_functions.h
+++ b/src/shared/common/vector_functions.h
@@ -105,6 +105,32 @@ template<> inline DeviceVec3d VecdZero() { return { static_cast<DeviceReal>(0.0)
 template<> inline Vec2d VecdZero() { return Vec2d::Zero(); }
 template<> inline Vec3d VecdZero() { return Vec3d::Zero(); }
 
+/* specialization for specific vector operations */
+template<class RealType, int Dimension>
+inline RealType VecdDot(const sycl::vec<RealType,Dimension>& v1, const sycl::vec<RealType,Dimension>& v2) {
+    return sycl::dot(v1, v2);
+}
+template<class RealType, int Dimension>
+inline RealType VecdDot(const Eigen::Matrix<RealType,Dimension,1>& v1, const Eigen::Matrix<RealType,Dimension,1>& v2) {
+    return v1.dot(v2);
+}
+template<class RealType, int Dimension>
+inline RealType VecdNorm(const sycl::vec<RealType,Dimension>& vec) {
+    return sycl::length(vec);
+}
+template<class RealType, int Dimension>
+inline RealType VecdNorm(const Eigen::Matrix<RealType,Dimension,1>& vec) {
+    return vec.norm();
+}
+template<class RealType, int Dimension>
+inline RealType VecdSquareNorm(const sycl::vec<RealType,Dimension>& vec) {
+    return sycl::dot(vec, vec);
+}
+template<class RealType, int Dimension>
+inline RealType VecdSquareNorm(const Eigen::Matrix<RealType,Dimension,1>& vec) {
+    return vec.squaredNorm();
+}
+
 /* SYCL memory transfer utilities */
 template<class T>
 inline T* allocateDeviceData(std::size_t size) {
--- a/src/shared/common/base_data_type.h
+++ b/src/shared/common/base_data_type.h
@@ -92,7 +92,7 @@ using AlignedBox3d = Eigen::AlignedBox<Real, 3>;
 using Rotation2d = Eigen::Rotation2D<Real>;
 using Rotation3d = Eigen::AngleAxis<Real>;
 /** Device data types. */
-using DeviceReal = float;
+using DeviceReal = Real;
 using DeviceVec2d = sycl::vec<DeviceReal, 2>;
 using DeviceVec3d = sycl::vec<DeviceReal, 3>;
 using DeviceArray2i = sycl::int2;
--- a/src/shared/kernels/kernel_wenland_c2.cpp
+++ b/src/shared/kernels/kernel_wenland_c2.cpp
@@ -92,7 +92,8 @@ DeviceReal DeviceKernelWendlandC2::W(const DeviceReal &r_ij, const DeviceVec3d &
 //=================================================================================================//
 DeviceReal DeviceKernelWendlandC2::W_1D(const DeviceReal q) const
 {
-    return sycl::pow(1.0f - 0.5f * q, 4.0f) * (1.0f + 2.0f * q);
+    return sycl::pow(static_cast<DeviceReal>(1.0) - static_cast<DeviceReal>(0.5) * q,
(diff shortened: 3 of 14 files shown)


ChiZhangatTUM:
Well done. I will go through your modification.

src/shared/common/vector_functions.h
```diff
@@ -105,6 +105,32 @@ template<> inline DeviceVec3d VecdZero() { return { static_cast<DeviceReal>(0.0)
...
+}
+template<class RealType, int Dimension>
+inline RealType VecdSquareNorm(const sycl::vec<RealType,Dimension>& vec) {
+    return sycl::dot(vec, vec);
+}
+template<class RealType, int Dimension>
+inline RealType VecdSquareNorm(const Eigen::Matrix<RealType,Dimension,1>& vec) {
+    return vec.squaredNorm();
+}
+
```
ChiZhangatTUM, line 133: I am curious that which operator is faster? for example sycl or eigen?

nR3D: As I was saying, a direct comparison is not possible since Eigen is not executable within SYCL kernels, at most it would be possible to compare Eigen running on CPU vs SYCL running either on GPU or CPU, but it would not be a fair comparison.

All I can say is that [SYCL geometric functions](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:geometric-functions) will map to GPU built-in intrinsic operations, which are supposedly faster than Eigen's. But Eigen offers more operations than SYCL. For instance, Eigen has already an optimized method for transposing a matrix while for SYCL you would need to implement one yourself, which would probably be slower compared to Eigen. Nevertheless, GPU processing power will probably still outweigh Eigen's optimizations on CPU.

Xiangyu-Hu:
@junwei-jiang Could you also have some comments?

Xiangyu-Hu:
the main slowdown is from
interval_computing_time_step =768.747579194 ?

junwei-jiang:
I'm curious about the precision difference. I mean, how much precision is lost when changing from double to float, and whether this is acceptable. I think the computational cost of double may be too high.

src/shared/common/vector_functions.h
ChiZhangatTUM: Did you notice info in this link? https://developer.codeplay.com/products/computecpp/ce/2.10.0/guides/eigen-guide/options-for-building-eigen 
Seems that eigen has a SYCL back-end.

nR3D:
> @Xiangyu-Hu
> the main slowdown is from
> interval_computing_time_step =768.747579194 ?

Since timestep computation and configuration updating are executed concurrently, that slowdown is probably determined by both. In general floating point numbers are used by all steps (aside from writing, where single/double precision is not relevant) so I would expect a similar slowdown for all methods.
I am now running the same computation again, separating timestep and configuration computations, so we can better distinguish the slowdown in different intervals. I will post the results as soon as they are ready.

> @junwei-jiang 
> I'm curious about the precision difference. I mean, how much precision is lost when changing from double to float, and whether this is acceptable. I think the computational cost of double may be too high.

Regression tests are still passing with floats, so double precision is not needed. But it is worth to have the possibility to enable it whenever an higher precision is required.
By default, device and host precision is set to be the same, so `SPHINXSYS_USE_FLOAT` may be toggled to use single precision for both.

Xiangyu-Hu:
> > @Xiangyu-Hu
> > the main slowdown is from
> > interval_computing_time_step =768.747579194 ?
> 
> Since timestep computation and configuration updating are executed concurrently, that slowdown is probably determined by both. In general floating point numbers are used by all steps (aside from writing, where single/double precision is not relevant) so I would expect a similar slowdown for all methods. I am now running the same computation again, separating timestep and configuration computations, so we can better distinguish the slowdown in different intervals. I will post the results as soon as they are ready.
> 
> > @junwei-jiang
> > I'm curious about the precision difference. I mean, how much precision is lost when changing from double to float, and whether this is acceptable. I think the computational cost of double may be too high.
> 
> Regression tests are still passing with floats, so double precision is not needed. But it is worth to have the possibility to enable it whenever an higher precision is required. By default, device and host precision is set to be the same, so `SPHINXSYS_USE_FLOAT` may be toggled to use single precision for both.

I would not worried about using float as the lead term to determine accuracy is numerical discretization algorithm other than double or single precision.

src/shared/common/vector_functions.h
nR3D: I did, but it seems that it's limited to tensor operations and, as far as I can tell, SYCL is used as a back-end for accelerating the execution of certain methods, not to support Eigen execution within SYCL kernels.

I am now having a specific look at what they upstreamed, but from a quick glance it seems that they re-implemented certain methods to be executed by SYCL, using raw pointers to allocated device memory. A similar approach to what I was proposing. Still, there might be some of those methods that may be reused.

ChiZhangatTUM: Thx for the info.

ChiZhangatTUM (approved):

nR3D:
Double:
```
Total wall time for computation: 2269.544049414 seconds.
interval_computing_time_step =46.734129965
interval_computing_fluid_pressure_relaxation = 653.920405645
interval_updating_configuration = 1458.222392889
interval_writing_files = 108.284276538
```

Float:
```
Total wall time for computation: 1376.905903631 seconds.
interval_computing_time_step =40.042162188
interval_computing_fluid_pressure_relaxation = 580.866289297
interval_updating_configuration = 610.196198424
interval_writing_files = 144.182650995
```

Benchmarks are still using 300'000 particles, but this time the server I am using to benchmark is partially busy so runtimes are slower than before, therefore I run both cases again to have a proper comparison.
Overall, configuration updating is where the execution slows down the most. I could run some profiling to find hotspots within configuration kernels. In particular, floating-point precision is relevant inside smoothing kernel calculations for neighbors, where different SYCL math and geometric functions are called, I suppose those are the reason for such slowdown.

PR #453, opened 2023-10-09
Draft PR: one-sided fluid-shell interaction

WeiyiKong321:
# Description
For one-sided fluid-shell interaction problems (e.g. blood flowing in elastic shell vessel), the shell particles cannot prevent penetration of fluid particles due to incompleteness of kernel support. 

# TODO
- Modifying dW_ijV_j and e_ij of shell neighbors by counting in the contribution of 'ghost particles' in the cut off radius of fluid particles

PR #465, opened 2023-11-06
Fluid shell interaction with contribution of ghost particles

WeiyiVirtonomy:
Same as PR #453 
https://github.com/Xiangyu-Hu/SPHinXsys/pull/453


WeiyiKong321:
# Observation: weird shell particle position

I made a test on Turek's fsi benchmark case. Under some conditions, the shell can have weird deformation:
image
image


The positions of shell particles are quite jaggy, and fluid particles fill in the spaces between two shell particles. I used several sets of parameters to see what caused this problem.

- resolution_fluid = 0.1, thickness = 0.2, resolution_shell = 0.1: occurs
- resolution_fluid = 0.1, thickness = 0.1, resolution_shell = 0.1: no
- resolution_fluid = 0.1, thickness = 0.1, resolution_shell = 0.05: no
- resolution_fluid = 0.05, thickness = 0.1, resolution_shell = 0.05: occurs

It seems that the problem might be related to the ratio between fluid spacing and shell thickness. It happens when thickness>fluid spacing.
@DongWuTUM Do you have any idea what might cause shell particles to have this weird distribution?

WeiyiKong321:
# Different shell resolution produces different results

Shell thickness = 0.1
- Shell spacing = thickness = 0.1
image

- Shell spacing = 0.5*thickness = 0.05
image

With the same thickness, when particle spacing decreases, the deformation of shell becomes much larger, and the vortex behind cylinder starts to detach earlier.

Xiangyu-Hu:
It looks like Hourglass mode.

DongWuTUM:
The results seem to have the instability of hourglass modes. However, it occurs only when the shell thickness is different from the fluid resolution. You can check if the contact force is correct or not because the hourglass modes may occur if the contact force is enormous or the distribution of force is extremely uneven. You can also try the hourglass control algorithm by  "Dynamics1Level<thin_structure_dynamics::ShellStressRelaxationFirstHalf> stress_relaxation_first_half(plate_body_inner, 3, true)“ referring to 2d plate case.

DongWuTUM:
Try to increase hourglass_control_factor_? But I don't think it is a good idea. Seems like quite large and uneven forces are the problems. Could we find a method to make the force smooth?

WeiyiKong321:
> Try to increase hourglass_control_factor_? But I don't think it is a good idea. Seems like quite large and uneven forces are the problems. Could we find a method to make the force smooth?

I increased the coefficient to 0.008. It seems that the hourglass modes are suppressed to a lower level, but as you said, the best way is to make sure fluid forces on the shell have a smoother distribution. Maybe we can try to average the fluid force in the inner neighbor of a shell particle, but I'm not sure if this is a good idea.

image

Xiangyu-Hu:
Please make sure the reason, or where is the exact problem, before new modification.  Because very often some issue is merely due to a bug.

WeiyiKong321:
The problem in 2D aortic valve mainly comes from U_max. It looks good after changing U_max to 2*U_f. I'll try to tune parameters of fsi2 as well
image

WeiyiKong321:
Looks like the 3d case can also work with a larger physical coefficient (I'll test with 3 leaflets tomorrow)
image

WeiyiKong321:
# Kernel completeness check
Fluid particles are relaxed using RelaxionStepComplex this time.
The initial $\sum_j W_{ij}V_j$ at this corner is too large (>1).
image

Very large $\sum_j dW_{ij}V_j\mathbf{e}_{ij}$:
image

Even for the cylinder, $\sum_j dW_{ij}V_j\mathbf{e}_{ij}$ is larger than that of the solid wall. The repelling force given by the shell boundary might be too strong. And I doubt the use of curvature, though volume is increased or decreased accordingly, $\sum_j dW_{ij}V_j\mathbf{e}_{ij}$ is almost the same as the one using constant volume.

# Hourglass of Turek's benchmark when thickness of shell>fluid_resolution
I tried to use damping or increase U_max, but it doesn't help. Sometimes using damping can make the condition even worse

WeiyiKong321:
It seems that the old formulation (W_ij = W_ij*thickness, dW_ijV_j=dW_ijV_j*thickness) performs better when shell thickness > resolution_fluid.
There are still fluid particles trapped inside shell at the beginning and the hourglass modes persist throughout the simulation, but most of the time fluid particles are kept away from the shell. What's more, when I use ghost particles, the displacement of the beam is always very small, while this one is closer to the volumetric solid results.
image
image

What I think now is that when shell_thickness > fluid_resolution, this simple integration gives a larger repelling force than the ghost particle method. The first one keeps fluid particles nearly 0.5*thickness away from the shell, while the second one can only keep them about 0.5 * resolution_fluid away.
Let's say thickness = 2 * dp_fluid=2 * dp_shell. The force of shell given by first formulation is F0 * dp * thickness=2 * F0 * dp * dp, while the one given by ghost particle integration is $F0 * dp * dp+F1 * dp * dp$, since the ghost particle is already 1dp more away from fluid particle, F1 will be much smaller than F0.
image
 
I still have no idea where the large ununiform force from fluid comes from.
image

For volumetric solid at the same time, the force is balanced from two sides. It looks like some shell particles receive force like the upper surface, some like the lower surface.
image

WeiyiKong321:
@Xiangyu-Hu I have updated the documentation. https://github.com/Xiangyu-Hu/SPHinXsys/tree/feature/one_sided_fluid_shell_interaction/tutorials/fluid_shell_interaction 
I would really appreciate it if you could check the first section when you have time.

DongWuTUM:
> @Xiangyu-Hu I have updated the documentation. https://github.com/Xiangyu-Hu/SPHinXsys/tree/feature/one_sided_fluid_shell_interaction/tutorials/fluid_shell_interaction I would really appreciate it if you could check the first section when you have time.

@WeiyiKong321 Weiyi, when previewing the 'one-side-fluid-shell-interaction.md' file on GitHub, some formulas are not displaying. Is there a good way to properly compile the Markdown file and ensure correct preview on GitHub?

WeiyiKong321:
> > @Xiangyu-Hu I have updated the documentation. https://github.com/Xiangyu-Hu/SPHinXsys/tree/feature/one_sided_fluid_shell_interaction/tutorials/fluid_shell_interaction I would really appreciate it if you could check the first section when you have time.
> 
> @WeiyiKong321 Weiyi, when previewing the 'one-side-fluid-shell-interaction.md' file on GitHub, some formulas are not displaying. Is there a good way to properly compile the Markdown file and ensure correct preview on GitHub?

There was no problem in the preview of vscode, not sure why they can't display properly on GitHub :( I've modified those formulas, could you please take a look again?

Xiangyu-Hu:
I have read the description and thinking that we should handle one-side and two side cases differently.
For one side, the integral should include all the ghost particle within cut-off radius.
However, for both side case, we should only take into account the ghost particles within the thickness, because there are still fluid particles on the other side of the shell.

WeiyiKong321:
> I have read the description and thinking that we should handle one-side and two side cases differently. For one side, the integral should include all the ghost particle within cut-off radius. However, for both side case, we should only take into account the ghost particles within the thickness, because there are still fluid particles on the other side of the shell.

I'll check how it works if it's integrated within the thickness. It's still quite different from the real shell geometry, which is from mid surface - 0.5*thickness to mid surface + 0.5*thickness, while the ghost particles are from mid surface to mid surface + 1*thickness

WeiyiKong321:
> I have read the description and thinking that we should handle one-side and two side cases differently. For one side, the integral should include all the ghost particle within cut-off radius. However, for both side case, we should only take into account the ghost particles within the thickness, because there are still fluid particles on the other side of the shell.

If it's integrated within the thickness, then maybe it's better to use Gaussian integration.

WeiyiKong321:
Shell self contact test
image

Xiangyu-Hu:
> Shell self contact test image

Great!

DongWuTUM:
> Shell self contact test image

Great. Thanks.

Xiangyu-Hu:
> # Some questions about coding style
> I'm not sure which one would be better, writing new classes for shell self-contact density summation and force, or just use `SelfContactDensitySummation` and `ContactDensitySummation`.
> 
> These two classes take `SurfaceContactRelation` as the input variable, if I want to reuse them, then `ShellSelfContactRelation` must be a child class of `SurfaceContactRelation`, which seems unnecessary. Also, as these classes use `Vol_` and `mass_`, I need to modify `dW_ijV_j` to `dW_ijV_j*particle_distance_*particle_distance_` in `NeighborBuilderShellSelfContact`. If you change those variables to `ParticleVolume` and `ParticleMass` someday, then `NeighborBuilderShellSelfContact` will also need modification, which might be inconvenient.

@WeiyiKong321 I would prefer not change the local dynamics but introduce new body relations. The reason is that we have so many different local dynamics and only handful body relations. Also the idea of the onside shell model is more on geometric other than a dynamic method.

WeiyiKong321:
> > # Some questions about coding style
> > I'm not sure which one would be better, writing new classes for shell self-contact density summation and force, or just use `SelfContactDensitySummation` and `ContactDensitySummation`.
> > These two classes take `SurfaceContactRelation` as the input variable, if I want to reuse them, then `ShellSelfContactRelation` must be a child class of `SurfaceContactRelation`, which seems unnecessary. Also, as these classes use `Vol_` and `mass_`, I need to modify `dW_ijV_j` to `dW_ijV_j*particle_distance_*particle_distance_` in `NeighborBuilderShellSelfContact`. If you change those variables to `ParticleVolume` and `ParticleMass` someday, then `NeighborBuilderShellSelfContact` will also need modification, which might be inconvenient.
> 
> @WeiyiKong321 I would prefer not change the local dynamics but introduce new body relations. The reason is that we have so many different local dynamics and only handful body relations. Also the idea of the onside shell model is more on geometric other than a dynamic method.

Thanks, I got it. May I change the dyanamic classes from `SelfContactDensitySummation(SelfSurfaceContactRelation &self_contact_relation)` to `SelfContactDensitySummation(BaseInnerRelation &self_contact_relation)`? I think it looks not so good to make `ShellSelfContactRelation` a child class of `SelfSurfaceContactRelation`, since there's no clear inheritance relation between them.

WeiyiKong321:
The method is not working when the shell is moving from two sides : ( I'll output the contact density and force, hope we can discuss tmr
image

Xiangyu-Hu:
> > > # Some questions about coding style
> > > I'm not sure which one would be better, writing new classes for shell self-contact density summation and force, or just use `SelfContactDensitySummation` and `ContactDensitySummation`.
> > > These two classes take `SurfaceContactRelation` as the input variable, if I want to reuse them, then `ShellSelfContactRelation` must be a child class of `SurfaceContactRelation`, which seems unnecessary. Also, as these classes use `Vol_` and `mass_`, I need to modify `dW_ijV_j` to `dW_ijV_j*particle_distance_*particle_distance_` in `NeighborBuilderShellSelfContact`. If you change those variables to `ParticleVolume` and `ParticleMass` someday, then `NeighborBuilderShellSelfContact` will also need modification, which might be inconvenient.
> > 
> > 
> > @WeiyiKong321 I would prefer not change the local dynamics but introduce new body relations. The reason is that we have so many different local dynamics and only handful body relations. Also the idea of the onside shell model is more on geometric other than a dynamic method.
> 
> Thanks, I got it. May I change the dyanamic classes from `SelfContactDensitySummation(SelfSurfaceContactRelation &self_contact_relation)` to `SelfContactDensitySummation(BaseInnerRelation &self_contact_relation)`? I think it looks not so good to make `ShellSelfContactRelation` a child class of `SelfSurfaceContactRelation`, since there's no clear inheritance relation between them.

Ok.

Xiangyu-Hu:
> # Hydrostatic fsi test
> I build the wall boundary also as shell, so that I can use the fluid damping to run this hydrostatic test.
> 
> ## shell_resolution=fluid_resolution=thickness
> When the thickness of elastic gate is 0.05, the maximum displacement is 1.0e-4, while the volumetric solid result 1.2e-4. When the thickness is 0.025, the maximum displacement is 6.7e-4, solid result is 7.8e-4. The force on shell from fluid is about 916. image
> 
> ## shell_resolution=fluid_resolution=0.5 * thickness
> However, when the shell resolution = 0.5 * thickness = 0.025, the maximum displacement becomes 1.7e-4, which is too large. The maximum force on shell from fluid is about 914.
> 
> ## shell_resolution=0.5 * fluid_resolution=0.5 * thickness
> This time the deformation becomes 1.1e-4. The force on shell from fluid is about 585.
> 
> It seems that the resolution and distance do have some influence on the result.

How far from the reference solution?

WeiyiKong321:
# Question about acc_prior caused by force from fluid

The pressure force on solid from fluid is:
```
Real face_wall_external_acceleration = (acc_prior_k[index_j] - acc_ave_[index_i]).dot(e_ij);
Real p_in_wall = p_k[index_j] + rho_n_k[index_j] * r_ij * SMAX(Real(0), face_wall_external_acceleration);
Real u_jump = 2.0 * (vel_k[index_j] - vel_ave_[index_i]).dot(n_[index_i]);
force += (riemann_solvers_k.DissipativePJump(u_jump) * n_[index_i] - (p_in_wall + p_k[index_j]) * e_ij) * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n];
```
For hydrostatic problem, `riemann_solvers_k.DissipativePJump(u_jump) * n_[index_i]` and  `rho_n_k[index_j] * r_ij * SMAX(Real(0), face_wall_external_acceleration)` are much smaller than `p_k[index_j]`. Neglecting these parts, the force is only related to `p_k[index_j]* e_ij * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n]`.

The pressure of fluid particles in contact with solid particles is all about rho_f * g * H, so the force is determined by `Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n] * e_ij `. The acceleration on the shell particle is then:

$$\frac{1}{\rho_s}\sum_jdW_{ij}V_j e_{ij}$$

Assume that shell particle distance dp_s is the same as fluid particle distance dp_f, then the relative distance between shell and fluid particles q=r_ij/h_f are almost the same.
image

This term will be proportional to inv_h of fluid, i.e. proportional to particle distance of fluid:

$$\sum_j dW_{ij}V_je_{ij} \propto \sum_j \frac{1}{dp_f^3}dW_1(q_{ij})dp_f^2e_{ij}=\sum_j \frac{1}{dp_f}dW_1(q_{ij})e_{ij}$$

Then the acceleration prior of the shell particle will be dependent on dp_f:
$$\frac{1}{\rho_s}\sum_jdW_{ij}V_j e_{ij} \propto \frac{1}{dp_f} \sum_j dW_1(q_{ij})e_{ij}$$

That's why when dp_f=dp_s=0.5*t=0.025, the displacement of shell becomes nearly twice that when dp_f=dp_s=0.05. Under the same conditions, the solutions change due to the changes in fluid resolution, which is surely wrong.

One way I can think of to fix the problem right now is to create a new kernel with a smoothing length of $dp_{avg} = 0.5*(dp_f+dp_s)$. The volume of a shell particle will be modified as $A_j * dp_{avg}$, instead of $A_j * t$, but the mass is still the actual mass $\rho_s * A_j * t$. Then the acceleration becomes:

$$\frac{V_i}{m_i}\sum_jdW_{ij}V_j e_{ij} \propto\frac{dp_{avg}}{rho_s*t}\frac{1}{dp_{avg}} \sum_j dW_1(q_{ij})e_{ij} \propto \frac{1}{t}$$

@DongWuTUM Dong, what do you think about this?

DongWuTUM:
> # Question about acc_prior caused by force from fluid
> The pressure force on solid from fluid is:
> 
> ```
> Real face_wall_external_acceleration = (acc_prior_k[index_j] - acc_ave_[index_i]).dot(e_ij);
> Real p_in_wall = p_k[index_j] + rho_n_k[index_j] * r_ij * SMAX(Real(0), face_wall_external_acceleration);
> Real u_jump = 2.0 * (vel_k[index_j] - vel_ave_[index_i]).dot(n_[index_i]);
> force += (riemann_solvers_k.DissipativePJump(u_jump) * n_[index_i] - (p_in_wall + p_k[index_j]) * e_ij) * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n];
> ```
> 
> For hydrostatic problem, `riemann_solvers_k.DissipativePJump(u_jump) * n_[index_i]` and `rho_n_k[index_j] * r_ij * SMAX(Real(0), face_wall_external_acceleration)` are much smaller than `p_k[index_j]`. Neglecting these parts, the force is only related to `p_k[index_j]* e_ij * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n]`.
> 
> The pressure of fluid particles in contact with solid particles is all about rho_f * g * H, so the force is determined by `Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n] * e_ij `. The acceleration on the shell particle is then:
> 
> 1ρs∑jdWijVjeij
> 
> Assume that shell particle distance dp_s is the same as fluid particle distance dp_f, then the relative distance between shell and fluid particles q=r_ij/h_f are almost the same. image
> 
> This term will be proportional to inv_h of fluid, i.e. proportional to particle distance of fluid:
> 
> ∑jdWijVjeij∝∑j1dpf3dW1(qij)dpf2eij=∑j1dpfdW1(qij)eij
> 
> Then the acceleration prior of the shell particle will be dependent on dp_f: 1ρs∑jdWijVjeij∝1dpf∑jdW1(qij)eij
> 
> That's why when dp_f=dp_s=0.5*t=0.025, the displacement of shell becomes nearly twice that when dp_f=dp_s=0.05. Under the same conditions, the solutions change due to the changes in fluid resolution, which is surely wrong.
> 
> One way I can think of to fix the problem right now is to create a new kernel with a smoothing length of dpavg=0.5∗(dpf+dps). The volume of a shell particle will be modified as Aj∗dpavg, instead of Aj∗t, but the mass is still the actual mass ρs∗Aj∗t. Then the acceleration becomes:
> 
> Vimi∑jdWijVjeij∝dpavgrhos∗t1dpavg∑jdW1(qij)eij∝1t
> 
> @DongWuTUM Dong, what do you think about this?

Have you tried the last equation? I have carefully read what you wrote. But I am still confused about "The force is determined by  `e_ij * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n]`." The force is also related to 'p_k[index_j]', right? When particle distance is changed to smaller, the 'p_k[index_j]' is also smaller. I may have misunderstood. If it is also convenient for you, we can have a discussion through Zoom.

DongWuTUM:
We can have a meeting now if it is Okay for you. The Zoom link is the same as before, and when you join, I will receive an email.

WeiyiKong321:
# Normal smoothing tested on T-shaped pipe

I made some changes to the one-sided NeighborBuilder. Now it will only calculate ghost particles when the normal direction of shell is in the opposite direction of e_ij, so particles in domain 1&3 cannot see the ghost particle of the vertical shell boundary, while particles in domain 2&3 cannot see ghost of horizontal shell boundary. Therefore, there's no more overlapping of ghost particles. For fluid particles in domain 3 there might be a risk of insufficient support, but it looks ok for this T-shaped pipe.
image

I tried to smoothen shell normal direction by the codes in ShellNormalPrediction:
`smoothed_[index_i] /= temp_[index_i].norm() + TinyReal;`

The results with and without normal smoothing doesn't show much difference. 

## No normal smoothing
image

## With normal smoothing
image

WeiyiKong321:
> > # Hydrostatic fsi test
> > I build the wall boundary also as shell, so that I can use the fluid damping to run this hydrostatic test.
> > ## shell_resolution=fluid_resolution=thickness
> > When the thickness of elastic gate is 0.05, the maximum displacement is 1.0e-4, while the volumetric solid result 1.2e-4. When the thickness is 0.025, the maximum displacement is 6.7e-4, solid result is 7.8e-4. The force on shell from fluid is about 916. image
> > ## shell_resolution=fluid_resolution=0.5 * thickness
> > However, when the shell resolution = 0.5 * thickness = 0.025, the maximum displacement becomes 1.7e-4, which is too large. The maximum force on shell from fluid is about 914.
> > ## shell_resolution=0.5 * fluid_resolution=0.5 * thickness
> > This time the deformation becomes 1.1e-4. The force on shell from fluid is about 585.
> > It seems that the resolution and distance do have some influence on the result.
> 
> How far from the reference solution?

I changed possion ratio to 0.49, so that we can compare to the analytical solution.

Analytical solution:
$$\delta_{max} = \frac{pL^4}{384EI}=7.3e-5$$

Shell with resolution_shell = resolution_fluid = shell_thickness = 0.05: 9e-5, about 23% error

Volumetric solid with resolution_shell = resolution_fluid = thickness/4 = 0.125: 6.9e-5, about 5% error

DongWuTUM:
> > > # Hydrostatic fsi test
> > > I build the wall boundary also as shell, so that I can use the fluid damping to run this hydrostatic test.
> > > ## shell_resolution=fluid_resolution=thickness
> > > When the thickness of elastic gate is 0.05, the maximum displacement is 1.0e-4, while the volumetric solid result 1.2e-4. When the thickness is 0.025, the maximum displacement is 6.7e-4, solid result is 7.8e-4. The force on shell from fluid is about 916. image
> > > ## shell_resolution=fluid_resolution=0.5 * thickness
> > > However, when the shell resolution = 0.5 * thickness = 0.025, the maximum displacement becomes 1.7e-4, which is too large. The maximum force on shell from fluid is about 914.
> > > ## shell_resolution=0.5 * fluid_resolution=0.5 * thickness
> > > This time the deformation becomes 1.1e-4. The force on shell from fluid is about 585.
> > > It seems that the resolution and distance do have some influence on the result.
> > 
> > 
> > How far from the reference solution?
> 
> I changed possion ratio to 0.49, so that we can compare to the analytical solution.
> 
> Analytical solution: δmax=pL4384EI=7.3e−5
> 
> Shell with resolution_shell = resolution_fluid = shell_thickness = 0.05: 9e-5, about 23% error
> 
> Volumetric solid with resolution_shell = resolution_fluid = thickness/4 = 0.125: 6.9e-5, about 5% error

The resolution of Fluid-shell interaction is smaller than Fluid-solid.

WeiyiKong321:
# Curvature change leads to leakage

Fluid particles tends to escape from the corner, possibly because the volume of ghost particles is reduced too much due to negative curvature. When adding ghost particle only for those satisfying displacement.dot(n_j) <= 0:
image
It's better when adding ghost particle for every shell particle
image

No curvature calculation, add ghost particle of each shell particle:
image

No curvature calculation, add ghost particle only for those satisfying displacement.dot(n_j) <= 0: tend to penetrate
image

I think we need to carefully choose a threshold close to one for the criterion displacement.dot(n_j) <= threshold

WeiyiKong321:
# Use ghost particles for shell-fluid contact

I think we should mimic the volumetric solid, and calculate the force from fluid on ghost shell particles as well. The force is then averaged over the thickness. The algorithm is similar to the fluid-shell contact: 

Assuming that the normal direction points from fluid to shell, the position of the ghost shell particle is:
$$r_{dummy} = r_i + n_i * dp_s$$ 
Then we calculate the summation of W_ij and e_ij:
$$\sum_{dummy} dW_{ij} V_j e_{ij}$$

The process is repeated until the distance between dummy shell particle and fluid particle j has exceeded the cut-off radius of fluid kernel.

The force on shell particle:
`force += (riemann_solvers_k.DissipativePJump(u_jump) * n_[index_i] - (p_in_wall + p_k[index_j]) * e_ij) * Vol_[index_i] * contact_neighborhood.dW_ijV_j_[n];`

Here, `Vol_[index_i]` the equal to shell particle area * dp_s. I think it would be better to use curvature information to calculate an equivalent volume of the shell, but I haven't tested it yet. 

The prior acceleration:
`acc_prior_[index_i] = force / particles_->ParticleMass(index_i);`

Here, the mass is the actual mass of shell, i.e. area * thickness. The total force from fluid is thus averaged over the shell thickness.

# Test results 
The method is tested on the hydrostatic fluid-shell interaction.
## dp_s < t (t=0.05, Dam_H = 2)
- dp_s=dp_f:
1. dp_s=dp_f=0.025: Analytical displacement: 0.000072667     Displacement: 0.000080300       Error: 10.505204766%
2. dp_s=dp_f=0.0125: Analytical displacement: 0.000072667     Displacement: 0.000073037       Error: 0.510098013%
3. dp_s=dp_f=0.0083: Analytical displacement: 0.000072667     Displacement: 0.000070717       Error: 2.682856359%
4. dp_s=dp_f=6.25e-3: Analytical displacement: 0.000072667     Displacement: 0.000069387       Error: 4.513355985%

- dp_s < dp_f:
5. dp_s=0.0125, dp_f=0.05 (1:4): Analytical displacement: 0.000072667     Displacement: 0.000072211       Error: 0.626392615%
6. dp_s=0.0125, dp_f=0.025 (1:2): Analytical displacement: 0.000072667     Displacement: 0.000073916       Error: 1.719810049%

## dp_s > t (t=0.0125, Dam_H = 1):
- dp_s=dp_f:
7. dp_s=dp_f=0.05(dp_s:t=4:1): Analytical displacement: 0.002325333     Displacement: 0.002797201       Error: 20.292465248%
8. dp_s=dp_f=0.025(dp_s:t=2:1): Analytical displacement: 0.002325333     Displacement: 0.002394119       Error: 2.958095353%

- dp_s < dp_f:
9. dp_s = 0.025, dp_f = 0.05: Analytical displacement: 0.002325333     Displacement: 0.002320483       Error: 0.208590640%

# Discussion
With ghost particles, the results are in the same range no matter what dp_s and dp_f we use, but there are still some problems.

If we look at case 1~4, the maximum displacement seems to be getting smaller as the resolution is refined, instead of converging to a constant value. One possible explanation is that the volume of shell particle is not corrected by curvature, so the total force becomes smaller when the layer of ghost particle increases. 

Another problem is that when the fluid particle gets too close to the shell particle in the initial step, it seems that fluid receives a large repelling force. In most cases the initial fluid particle relaxation should not be a problem, and this seems more likely to be caused by the instability of inviscid fluid, instead of fluid-shell contact.
image
image

I'm not sure if this method makes sense. I hope we can have some further discussion tmr.

WeiyiKong321:
# Suggestion

I think it would be better to change the ParticleMass() to mass_ in `BasePressureForceAccelerationFromFluid`:

`Real face_wall_external_acceleration = (force_prior_k[index_j] / mass_k[index_j] - force_ave_[index_i] / particles_->ParticleMass(index_i)).dot(e_ij);`

`force_ave_` is defined as the area force for shell, so the acc_ave_ of shell should be `force_ave_[index_i] / particles_->mass_[index_i]`.

If you change it to `particles_->mass_[index_i]`, I can reuse classes in `fluid_structure_interaction.h` for force from fluid on shell, instead of writing new classes.

Xiangyu-Hu:
> # Suggestion
> I think it would be better to change the ParticleMass() to mass_ in `BasePressureForceAccelerationFromFluid`:
> 
> `Real face_wall_external_acceleration = (force_prior_k[index_j] / mass_k[index_j] - force_ave_[index_i] / particles_->ParticleMass(index_i)).dot(e_ij);`
> 
> `force_ave_` is defined as the area force for shell, so the acc_ave_ of shell should be `force_ave_[index_i] / particles_->mass_[index_i]`.
> 
> If you change it to `particles_->mass_[index_i]`, I can reuse classes in `fluid_structure_interaction.h` for force from fluid on shell, instead of writing new classes.

Actually, I prefer to use the particle mass function because it is misleading. Anyway, we can discuss it tommorow.

WeiyiKong321:
# About the calculation of shell curvature using the cut-off radius of fluid

I'm not sure what would be the most efficient way to do this. Should I create a new CellLinkedList of shell with the cut-off radius of fluid, or try to toggle with the SearchDepth?

DongWuTUM:
> # About the calculation of shell curvature using the cut-off radius of fluid
> I'm not sure what would be the most efficient way to do this. Should I create a new CellLinkedList of shell with the cut-off radius of fluid, or try to use the same CellLinkedList but changing the SearchDepth?

If the cut-off radius is changed, the CellLinkedList should be changed accordingly.

WeiyiKong321:
> > # About the calculation of shell curvature using the cut-off radius of fluid
> > I'm not sure what would be the most efficient way to do this. Should I create a new CellLinkedList of shell with the cut-off radius of fluid, or try to use the same CellLinkedList but changing the SearchDepth?
> 
> If the cut-off radius is changed, the CellLinkedList should be changed accordingly.

I'm a bit confused about this. Can we have a zoom meeting?

WeiyiKong321:
The curvature calculated from fluid cut-off radius looks good at the 90 corner.
image

Without using B_matrix, the curvature of particles near the corner is quite close, and the corner curvature is slightly smaller than the 2 particles beside it. This doesn't cause problem for fluid flow, but not sure if the inaccuracy will cause problem at sharper corner.
image

If we want to use B matrix correction for curvature calculation of an elastic shell, then a new kind of B must be registered. I'll continue to test after back from vacation.

WeiyiKong321:
# Test on flow around a flat balloon

Fluid resolution dp_f = 0.15, shell resolution dp_s = 0.075. The disthe upper and lower layers is 1 * dp_f.

image

We can't use correction matrix B for this case, otherwise the curvature of the flat part will be calculated to 0:
image

When B matrix is not used, the curvature of the upper and lower layer is 9.1 and 10.1 respectively.
 
Theoretically, there should be 1 ghost particle in total:
image

The length of the 1st ghost particle generated by the upper shell particle is (1-9.1 * 0.075) * dp_s = 0.32 * dp_s, while the length of the 2nd is already below 0. That generated by the lower shell particle is (1-10.1 * 0.075) * dp_s = 0.24 * dp_s. Thus, the length of the ghost particle generated in total is 0.56 * dp_s, which is smaller than the theoretical value of 1 * dp_s.
 
Although some fluid particles seem to be too close to the shell, the fluid field looks still ok.
image

image

I haven't tested on smaller distances since there will be some problems with normal direction, but we can still check the curvature:
image

The distance is now 1.02 * dp_s. The curvature of the flat part is 10.8 and 9.7, which doesn't differ much from the 2 * dp_s case, while there shouldn't be any ghost particle in this case.

It seems that the curvature is not accurate enough when B correction is not used, but when B is used, it becomes the real curvature.

WeiyiKong321:
# Test on flow around an elastic balloon

An elastic balloon is placed in the flow. The inlet velocity gradually increases to the maximum value from 0 to 1 sec, and maintains constant thereafter. The balloon is stiff enough that no self-contact occurs.
image

At 0.5s:
image

At 1s, the curvature correction is working properly:
image

At 1.5s:
image

At 2.5s:
image

The balloon nearly recovers to the initial shape when the flow becomes stable:
image

WeiyiKong321:
# Discussion: use different mean curvature for shell self-contact and fluid-shell interaction

For shell self-contact problem, the curvature must be the actual curvature with B matrix correction. I would suggest to use two separate variables "TotalMeanCurvature" and "AverageTotalMeanCurvature" for self-contact and fluid-shell interaction, and two different classes to calculate them. What do you think about this?

DongWuTUM:
Thank you! Then, you could try not to use the B for curvature calculation. The results seem Okay. For the figure below, how do you get the normal direction, and could you try to get the correct one first? Or you could generate the particle by directly giving the particle position and normal direction.
image

The results of the test on flow around an elastic balloon are very good. I suggest no adjustment on the volume of ghost particles for self-contact problem. But could you try both and check which one is better?

WeiyiVirtonomy:
# Fail on 3D pipe with sharp angle

The pipe has a sharp angle of 10 degree. Fluid resolution = radius / 20 = 0.03, shell resolution = 0.5 * fluid resolution = 0.015.
image

Very high velocity singularities near the corner, though the total kernel is not higher than the other particles. There's also no obvious particle repelling.
image

I'll try to add a fillet at the corner and see if the result will be better.

WeiyiVirtonomy:
> Thank you! Then, you could try not to use the B for curvature calculation. The results seem Okay. For the figure below, how do you get the normal direction, and could you try to get the correct one first? Or you could generate the particle by directly giving the particle position and normal direction. image
> 
> The results of the test on flow around an elastic balloon are very good. I suggest no adjustment on the volume of ghost particles for self-contact problem. But could you try both and check which one is better?

I was using shell particle relaxation from SPHinXsys. It's a bit difficult to assign normal for this case, I'll try on a 3D case later and use virtonomy's surface remeshing to get the normal.

I haven't seen much difference in self-contact with or without curvature change yet, but not sure what it will be like when the deformed shape has very large curvature.

DongWuTUM:
> # Fail on 3D pipe with sharp angle
> The pipe has a sharp angle of 10 degree. Fluid resolution = radius / 20 = 0.03, shell resolution = 0.5 * fluid resolution = 0.015. image
> 
> Very high velocity singularities near the corner, though the total kernel is not higher than the other particles. There's also no obvious particle repelling. image
> 
> I'll try to add a fillet at the corner and see if the result will be better.

At the corner, there should be high-velocity particles. You are right; the sharp corner may cause problems (singularity).

WeiyiVirtonomy:
> > # Fail on 3D pipe with sharp angle
> > The pipe has a sharp angle of 10 degree. Fluid resolution = radius / 20 = 0.03, shell resolution = 0.5 * fluid resolution = 0.015. image
> > Very high velocity singularities near the corner, though the total kernel is not higher than the other particles. There's also no obvious particle repelling. image
> > I'll try to add a fillet at the corner and see if the result will be better.
> 
> At the corner, there should be high-velocity particles. You are right; the sharp corner may cause problems (singularity).

dp_f = 0.4, dp_s = 0.2:
# Fillet of radius 0.2
Compared with solid wall:
image

At the same time, unlike solid wall, shell wall still has velocity near the corner:

Shell:
image

Solid:
image

At 10s, the singularities seem to disappear:

Shell:
image

Solid:
image

# Fillet of radius 0.3
With a fillet of radius 0.3, the result is closer to that of solid:
image

image

Comparison of wall particle distribution at the corner:
image

I still don't know if the singularities are caused by geometry or shell ghost particles. Since W_ijV_j summation near corner seems to be normal, I think it's more likely to be caused by geometry.

DongWuTUM:
It seems Okay when the fillet is applied on the sharp corner. I think it is the geometry problem.

DongWuTUM:
The density has been updated. You can use it to update the volume. But I don't know if it is necessary to do so.

Xiangyu-Hu:
> # About the volume/area of shell
> We are not updating the volume/area of shell during deformation, but for problems like inflation of an elastic balloon, the volume of shell can have considerable changes. I think it might be necessary to update the volumes of shell by mass / density in those cases, so that the fluid-shell contact relation can update with the current volume of neighboring shell particles.

@WeiyiVirtonomy As we using Total Lagrangian for solid dynamics, in which the formulation is only dependent on in initial configuration, we need  not use particle  volume/area there. If we need to update it, I prefer do not update it in contact relation but in a FSI method, such as in AverageVelocityAndAcceleration. Certainly, in this case the method should be renamed properly.

WeiyiVirtonomy:
# Using principle curvatures for 3D problems

For 3D problems, the mean curvature is not accurate enough. We should use 2 principle curvatures to calculate the volume change of ghost particles.

Test on 3D pipe with sharp corner: no singularities near the corner even without using fillet
image

Test on flow around flat plate:
Screenshot 2024-01-11 171344

Xiangyu-Hu:
> # Using principle curvatures for 3D problems
> For 3D problems, the mean curvature is not accurate enough. We should use 2 principle curvatures to calculate the volume change of ghost particles.
> 
> Test on 3D pipe with sharp corner: no singularities near the corner even without using fillet image
> 
> Test on flow around flat plate: Screenshot 2024-01-11 171344

@WeiyiVirtonomy Nice work!

WeiyiVirtonomy:
# Test on 3D fluid-shell interaction

Flow around an elastic balloon:
image

It's still not very steady for 3d fsi. I have to take a physical length 10 to 20 times of the empirical value, otherwise shell will break very quick.
image

Also noticed wrong velocity near the shell in some simulations. Maybe principle curvature estimation is still not accurate enough for unsymmetric shell shape
image

WeiyiVirtonomy:
Is it correct to only use the velocity of a shell particle when calculating the force from shell on fluid? Shall we consider the angular velocity of shell?

FabienPean-Virtonomy:
> Is it correct to only use the velocity of a shell particle when calculating the force from shell on fluid? Shall we consider the angular velocity of shell?

It depends on 2 things in my understanding:
1. is the contact between fluid and shell only at the mid-surface or the actual half-thickness offset location
2. is it a no-slip condition between fluid and the shell

I haven't followed recent technical choices, but if both answers are yes, then the effective velocity at integration location is most likely needed.

WeiyiVirtonomy:
> > Is it correct to only use the velocity of a shell particle when calculating the force from shell on fluid? Shall we consider the angular velocity of shell?
> 
> It depends on 2 things in my understanding:
> 
> 1. is the contact between fluid and shell only at the mid-surface or the actual half-thickness offset location
> 2. is it a no-slip condition between fluid and the shell
> 
> I haven't followed recent technical choices, but if both answers are yes, then the effective velocity at integration location is most likely needed.

Sounds reasonable. I think the velocity at mid-surface needs to be transformed to the velocity at half-thickness or half-dp_s offset, or maybe some more accurate effective velocity.
@DongWuTUM Do you have some time this week to discuss this issue? I saw the strange void in 2D cases as well, maybe related to the wrong shell velocity.
image

DongWuTUM:
Actually, we have used the angular velocity as we use the curvature to calculate the volume of dummy particles. The method that we use now is very similar to fluid-solid contact. 

We can have a discussion tomorrow at 2 p.m. if that works for you.

WeiyiVirtonomy:
> Actually, we have used the angular velocity as we use the curvature to calculate the volume of dummy particles. The method that we use now is very similar to fluid-solid contact.
> 
> We can have a discussion tomorrow (Jan. 16) at 2 p.m. if that works for you.

That works for me. Thank you : )

WeiyiVirtonomy:
# Test on a moving shell-boundary

The balloon is pushed by external force, with a period of 0.1s. Fluid force on balloon is not considered, i.e. the balloon acts as a moving rigid wall to the fluid. Inlet average velocity U_f = 0.28m/s, Re=1000. The velocity of shell particle in contact can reach 0.9m/s.

dp_f=0.6mm, dp_s=0.3mm:
image
image

dp_f=0.3mm, dp_s=0.3mm:
(Fluid breaks behind the balloon after several cycles, since some particles are wrongly identified as surface particles.)
image
image

The velocity distribution near the balloon seems good. The pressure in the flow field is more messy, but it's normal for SPH. I think the force from shell on fluid is ok, the problem may come from fluid force on shell. Shell can also be more instable than solid, so the messy fluid pressure around may cause the shell to fail.

When testing on the pumping of balloon, I also noticed that the hourglass effects can occur when the force is too large. I'll check the classes of shell self-contact again.

DongWuTUM:
Thank you. Are you using the periodic inflow condition? If you are, surface particles do not need to be identified. Otherwise, there are void areas on the right-hand side.

Xiangyu-Hu:
@WeiyiVirtonomy as the balloon volume changes a lot, you may not use periodic boundary condition as it does not change the total number of particles and leads to particle volume change against the weakly compressible assumption.

WeiyiVirtonomy:
> Thank you. Are you using the periodic inflow condition? If you are, surface particles do not need to be identified. Otherwise, there are void areas on the right-hand side.

I'm using emitter inlet and deletion outlet, but I think this problem is not relevant to fsi

WeiyiVirtonomy:
I wanted to reproduce the results in this paper Chapter 12.3: https://www.researchgate.net/profile/Tayfun-Tezduyar/publication/229866002_Modeling_of_fluid-structure_interactions_with_the_space-time_finite_elements_Solution_techniques/links/59dfc9500f7e9bc51256c42c/Modeling-of-fluid-structure-interactions-with-the-space-time-finite-elements-Solution-techniques.pdf
https://www.tafsm.org/PROJ/FSI/Balloon/

image
Pumping air into an balloon. 
-Initial radius of balloon: 2m
-Diameter of circular hole: 0.6245m
-Thickness of balloon: 2e-3m
-Young's modulus of balloon: 1e3 N/m^2
-Density of fluid: 1 kg/m^3
-Viscosity of fluid: 1.5e-5 m2/s
-Inlet velocity: cosine wave, period=2s, minimum vel=0, maximum vel=2m/s
The paper didn't mention the Possion ratio and material type of balloon, so I used Saint Venant solid with a possion=0.495. The whole inlet pipe is assigned with a uniform velocity.
In the paper it seems like the balloon is empty initially, but I filled it with air from the beginning. Maybe the way I set the inlet cannot reproduce the pumping condition, the deformation pattern is completely different from the results in the paper.

At 2s:
image

At 4s:
image

WeiyiVirtonomy:
I wanted to reproduce the 2d case of fluid filling into an elastic container. The viscosity of fluid is very high, up to 50 Pa.s
https://www.researchgate.net/publication/318146847_A_partitioned_fully_explicit_Lagrangian_Finite_Element_Method_for_highly_nonlinear_Fluid-Structure-Interaction_problems
image

However, this case is very unstable even with volumetric solid. Once there's contact of fluid and solid, the deformation of solid becomes very large and soon breaks.
image

I have several questions regarding this kind of problem:
1. Should I use free stream of free surface density summation?
2. Should I use transport velocity correction?
3. The paper mentioned the bulk modulus of fluid, should I choose sound speed of fluid based on sqrt(K*rho_0)?
4. Would it be better if we execute every thing in a solid time step size, instead of using a tri-loop?

Xiangyu-Hu:
> I wanted to reproduce the 2d case of fluid filling into an elastic container. The viscosity of fluid is very high, up to 50 Pa.s https://www.researchgate.net/publication/318146847_A_partitioned_fully_explicit_Lagrangian_Finite_Element_Method_for_highly_nonlinear_Fluid-Structure-Interaction_problems image
> 
> However, this case is very unstable even with volumetric solid. Once there's contact of fluid and solid, the deformation of solid becomes very large and soon breaks. image
> 
> I have several questions regarding this kind of problem:
> 
> 1. Should I use free stream of free surface density summation?
> 2. Should I use transport velocity correction?
> 3. The paper mentioned the bulk modulus of fluid, should I choose sound speed of fluid based on sqrt(K/rho_0)?
> 4. Would it be better if we execute every thing in a solid time step size, instead of using a tri-loop?

> I wanted to reproduce the 2d case of fluid filling into an elastic container. The viscosity of fluid is very high, up to 50 Pa.s https://www.researchgate.net/publication/318146847_A_partitioned_fully_explicit_Lagrangian_Finite_Element_Method_for_highly_nonlinear_Fluid-Structure-Interaction_problems image
> 
> However, this case is very unstable even with volumetric solid. Once there's contact of fluid and solid, the deformation of solid becomes very large and soon breaks. image
> 
> I have several questions regarding this kind of problem:
> 
> 1. Should I use free stream of free surface density summation?
> 2. Should I use transport velocity correction?
> 3. The paper mentioned the bulk modulus of fluid, should I choose sound speed of fluid based on sqrt(K/rho_0)?
> 4. Would it be better if we execute every thing in a solid time step size, instead of using a tri-loop?

Use free surface density summation. No transport, fluid sound speed is from bulk modulus, you need explore the time stepping as I do not have experience on this either.

WeiyiVirtonomy:
> > I wanted to reproduce the 2d case of fluid filling into an elastic container. The viscosity of fluid is very high, up to 50 Pa.s https://www.researchgate.net/publication/318146847_A_partitioned_fully_explicit_Lagrangian_Finite_Element_Method_for_highly_nonlinear_Fluid-Structure-Interaction_problems image
> > However, this case is very unstable even with volumetric solid. Once there's contact of fluid and solid, the deformation of solid becomes very large and soon breaks. image
> > I have several questions regarding this kind of problem:
> > 
> > 1. Should I use free stream of free surface density summation?
> > 2. Should I use transport velocity correction?
> > 3. The paper mentioned the bulk modulus of fluid, should I choose sound speed of fluid based on sqrt(K/rho_0)?
> > 4. Would it be better if we execute every thing in a solid time step size, instead of using a tri-loop?
> 
> > I wanted to reproduce the 2d case of fluid filling into an elastic container. The viscosity of fluid is very high, up to 50 Pa.s https://www.researchgate.net/publication/318146847_A_partitioned_fully_explicit_Lagrangian_Finite_Element_Method_for_highly_nonlinear_Fluid-Structure-Interaction_problems image
> > However, this case is very unstable even with volumetric solid. Once there's contact of fluid and solid, the deformation of solid becomes very large and soon breaks. image
> > I have several questions regarding this kind of problem:
> > 
> > 1. Should I use free stream of free surface density summation?
> > 2. Should I use transport velocity correction?
> > 3. The paper mentioned the bulk modulus of fluid, should I choose sound speed of fluid based on sqrt(K/rho_0)?
> > 4. Would it be better if we execute every thing in a solid time step size, instead of using a tri-loop?
> 
> Use free surface density summation. No transport, fluid sound speed is from bulk modulus, you need explore the time stepping as I do not have experience on this either.

Thank you for the kind reply. I tried to use min(dt, dt_s) as the single time step size, and these are the results:

- t = 0.92 s
image

- t = 1.04 s
image

- t = 1.20 s
image

- t = 1.50 s
image

- t = 2.00 s
image

- t = 2.80 s
image

- t = 3.50 s
image

- t = 3.80 s
image

- t - 10.0s
image

The deformation at 3.5 and 3.8 s is quite different from the paper, others are similar. I'll try to increase resolution and compare the displacement curve.

I used a quite larger damping for shell (100 * 0.4 / 4.0 * std::sqrt(rho0_s * youngs_modulus) * s;), will also check if this value can be reduced.

DongWuTUM:
It is a dynamic case, right? Why are you using the damping which is for steady state?

WeiyiVirtonomy:
> It is a dynamic case, right? Why are you using the damping which is for steady state?

I have to use damping to stabilize the solid / shell, though it will cause loss of accuracy

Xiangyu-Hu:
> > It is a dynamic case, right? Why are you using the damping which is for steady state?
> 
> I have to use damping to stabilize the solid / shell, though it will cause loss of accuracy. I'm testing on switching off damping now

Seems like the issue added mass.

WeiyiVirtonomy:
I don't know why, hourglass effects seem to be more serious when using AllForceFromFluidRiemann

This is the result with or without Riemann at the same time step:

With Riemann:
image

Without Riemann:
image

WeiyiVirtonomy:
# Verification case: elastic gate
t=0.04s
0_04s

t=0.08s
0_08s

t=0.12s
0_12s

t=0.16s
0_16s

Without using damping, there's still hourglass
image

Comparison of x and y displacement at the bottom of gate:
Reference: https://www.researchgate.net/publication/325787272_An_enhanced_ISPH-SPH_coupled_method_for_simulation_of_incompressible_fluid-elastic_structure_interactions

V_disp
H_disp

WeiyiVirtonomy:
# Hydrostatic case
I think the parameters of this benchmark case should be the same with hydrostatic_fsi, but even the displacement from volumetric solid is 1.3e-4, while the analytical solution is 6.85e-5.

WeiyiVirtonomy:
# Flow around cylinder

Haven't done convergence test yet, it's only a comparison of total pressure & viscous force on the cylinder for solid and shell. Fluid resolution and solid/shell resolution are the same.

Total pressure force:
force_p

Total viscous force:
force_v

The error of average total pressure force and viscous force is 11.7% and 14.2% respectively.

Part of the difference is caused by the fact that I didn't modify r_ij of the shell particles. If directly adding ghost neighbors into the Neighborhood (to eliminate the effect of wrong r_ij), the error becomes 8.3% and 6.4%. 

@DongWuTUM Could you please also send me a reference paper from SPHinXsys with the force on cylinder? Some other references seem to have a smooth force curve

DongWuTUM:
Hi Weiyi, Thank you for the testing. I have sent you an email including the flow around the cylinder and other fluid-shell interaction cases suitable for the test. See details and comments in the email.

For comparison between solid and shell, the same resolution doesn't mean the results should be the same. I mean the comparison should be conducted when the results of both are converged.

WeiyiVirtonomy:
Dear Dong,

Thank you so much for providing me with abundant materials. I'll try those cases according to your comments.

Btw, do we have any cases of dynamic shell without using damping? It seems that shell still has hourglass modes without using damping, even for very simple cases.

Best regards,
Weiyi Kong

DongWuTUM:
Hi Weiyi, you could refer to the test case, dw_3d_dynamic_plate. You could also check my paper in arxiv, Wu, D., Zhang, C., & Hu, X. (2023). An SPH formulation for general plate and shell structures with finite deformation and large rotation. arXiv preprint arXiv:2309.02838.

WeiyiVirtonomy:
Thanks a lot. If the dynamic plate works without damping, then the instability in FSI seems to be coming from fluid forces on shell, instead of shell itself. The curvature-ghost particle method may be still not accurate enough.

WeiyiVirtonomy:
# 3D benchmark: dambreak with an elastic plate
Reference: https://www.researchgate.net/publication/348498997_An_accurate_FSI-SPH_modeling_of_challenging_fluid-structure_interaction_problems_in_two_and_three_dimensions

Initial H0=2m
- t=0.27s
0_27s

- t = 0.32s
0_32s

- t = 0.37s
0_37s

- t = 0.42s
0_42s

- t = 0.47s
0_47s

Horizontal displacement of the observer (not converged yet, but the displacement is close to each other under different resolutions):
disp

WeiyiVirtonomy:
@DongWuTUM Hi, Dong. May I ask some questions about the wedge impacting?
I was trying to reproduce the test case in this paper:
Reference: https://www.researchgate.net/publication/325787272_An_enhanced_ISPH-SPH_coupled_method_for_simulation_of_incompressible_fluid-elastic_structure_interactions
(Figure 20)

The paper didn't mention the density of wedge, the material properties of water and the size of water domain. Since the wedge is aluminum, I set density as 2700kg/m3. The parameters of water are the same as that composite wedge paper: half length = 2.5m, height = 1.5m, density 1000kg.m3, invisid.

According to the paper, the wedge's DOF is fixed at the center and the extremities, so I set angular velocity to (0,0) and velocity to (0,-30) at point A, and the left and right end of the wedge.

Initial geometry:
image

The problem is, I have very weird deflection at point C (calculated by -disp - 30 * time), completely different from the reference:
disp

Pressure at point A has instability:
pA

Wedge has hourglass modes, and fluid particles seem to be squeezed:
image

If I set an initial gap between fluid and wedge, then wedge has hourglass modes before it interacts with water
image

I have several questions about this case:
1. Should we use damping for fluid, like in the hydrostatic fsi case?
2. How to suppress the hourglass of wedge?
3. The end time is about 2e-3s, under dp_f = 0.02 and U_max = 30*2m/s, Dt_ref = 0.0001, which means that configuration is only updated for 20 times during the simulation. Do we need to use larger U_max and c_f?
4. I'm not sure if I understand the conditions in the reference correctly

Could we have a call to discuss this a bit when you have time? Thanks a lot

DongWuTUM:
@WeiyiVirtonomy Hi Weiyi, we can have a discussion at 3 p.m., if it is also Okay for you.

WeiyiVirtonomy:
Thanks. I have another meeting at 3 p.m. today, and I'm not sure how much time it will take. Is there any other time suitable for you today? Or we can also discuss after tomorrow's meeting

DongWuTUM:
OK, let us discuss after tomorrow meeting.

WeiyiVirtonomy:
# Update: 2d version of the 3d dambreak

2D has stronger hourglass modes than the 3D case near the constraint:
image

The hourglass modes remain even after using physical damping:
image

Tip displacements under different resolutions:
disp

Comparison with 3d:
disp

WeiyiVirtonomy:
# Wedge impact:
If the mid point is not fixed, there's no hourglass, but the displacement seems to be larger than the semi-analytical solution.
image

When using larger U_max (200*30m/s) and c_f for fluid, there's a larger gap between fluid particles:
image
image

I'm not sure if this comes from the instability of fluid.

WeiyiVirtonomy:
# Wedge impacting free surface
Ref: https://doi.org/10.1016/j.cpc.2018.05.012

Results only fixing extremities & using single time resolution:

At 0.0009s:
image

At 0.0017s:
image

At 0.0025s:
image

Displacement:
disp

Pressure at A:
pA

Pressure at D:
pD

The pressure seems to be lower than the semi-analytical solution. At point D, where there seems to be a delay of the shell-fluid contact. This also can be seen from the screenshots, where the height of water surface at the last time step seems to be lower than that in the reference paper.

The results seem to have a relatively large deviation from the paper, I'll delete it if you don't need this test case.

@DongWuTUM As for the case of fluid filling elastic container, I haven't checked how to use SimTK::MultibodySystem yet, so I don't know how much time it will take to add wetting to it. Could we just use the current results?

DongWuTUM:
Hi, Weiyi. We can ignore the wedge case now. For the case of fluid filling elastic container, we don't need the Simbody algorithms. 
The wetting effect just serves as a boundary condition. You can check the lines 330 - 334 of the `test_2d_water_entry_exit`. If you don't have time to try, we can use the current results.

WeiyiVirtonomy:
> Hi, Weiyi. We can ignore the wedge case now. For the case of fluid filling elastic container, we don't need the Simbody algorithms. The wetting effect just serves as a boundary condition. You can check the lines 330 - 334 of the `test_2d_water_entry_exit`. If you don't have time to try, we can use the current results.

Looks like the Dirichlet condition of phi is fulfilled, but the displacement doesn't have much difference
image

Initial phi of water =1, initial phi of container = 0

DongWuTUM:
Okay, thanks :) It also makes sense that the difference is negligible in this case.

WeiyiVirtonomy:
For contact problems, I'm not sure if curvature correction is necessary. I'll try with constant volume first, and check if there will be problem when the body has contact with a shell in self contact:
image

Also, for shell contact force, a limitation coefficient might be needed. I'm wondering if I should add a coefficient in `RepulsionForce<Contact<Wall>>`, or adjust the force by scaling dW_ije_ij in neighbor builders.

DongWuTUM:
From the idea of applying dummy particles, there should be curvature correction. With curvature correction, a limitation coefficient might be not needed. You could have a try, and then we could have a discussion based on the test results.

WeiyiVirtonomy:
# A question about class `ContactDensitySummation`

From what I understand, if a solid body 1 is interacting with a rigid solid body 2, the density of body 2 should be the same as body 1. Is this correct?
Have you considered writing a new density summation class for solid-rigid body interaction, which scales the contact mass by the density of the body itself?

WeiyiVirtonomy:
# 2d sliding: test on solid-rigid shell contact
Ref: DOI:[10.1016/j.jcp.2022.111105](http://dx.doi.org/10.1016/j.jcp.2022.111105)

image

Center displacement:
disp

If the shell has two layer like the self-contact condition, the result is similar:
image

At the end the solid seems to be quite unstable, but this might be caused by the relatively small E=1e5. 
image

DongWuTUM:
> # A question about class `ContactDensitySummation`
> From what I understand, if a solid body 1 is interacting with a rigid solid body 2, the density of body 2 should be the same as body 1. Is this correct? Have you considered writing a new density summation class for solid-rigid body interaction, which scales the contact mass by the density of the body itself?

Hi Weiyi. Don't need to be same. You can check the test cases, such as `test_2d_ball_shell_collision`.

WeiyiVirtonomy:
# About hourglass control coefficient

I have increased the coefficient to 2, i.e. 1000 times of the default value, but the hourglass modes still exist
image

WeiyiVirtonomy:
# Flow around balloon initial failing

No inlet velocity, the fluctuation of pressure comes from density summation.
dp_s=0.3 mm, dp_f= 0.6 mm, thickness =0.3mm, E=3.4e3

Pressure force from fluid no Riemann:
image

Pressure force from fluid Riemann:
image

Refine fluid resolution:
dp_s=0.3 mm, dp_f= 0.3 mm

Pressure force from fluid no Riemann:
image

Pressure force from fluid Riemann:
image

No matter whether the shell breaks or not, using PressureFromFluidRiemann leads to balloon weirdly inflating.

WeiyiVirtonomy:
@DongWuTUM Hi, Dong, could you tell me again which test case can be used to study shell instability under external forces?

DongWuTUM:
For the problem according to "PressureFromFluidRiemann leads to balloon weirdly inflating," we should discuss it with Prof. Hu. 
I think Prof. Hu means using the class `SurfacePressureFromSource` in `loading_dynamics.h` to apply pressure. But I notice that there is a hard coding in `Real area = pow(particles_->Vol_[index_i], 2.0 / 3.0)`. You should change it for 2D simulations.

WeiyiVirtonomy:
# Test on beam with external load

The geometry and material properties of the beam is the same as in test case 2d_elastic_gate.
The force mimicks this time step (0.028s) in fsi, where stress zigzag can be seen near the constraint: 
image
I apply a force that increases linearly at the fixed part and remains constant at the moving part. The maximum force is 0.5. When t < 0.028s, force linearly increases to the target value, then keeps constant thereafter.
image

The beam starts to oscillate under such external pressure force.

Stress hourglass near the constraint part:
image

Even stronger hourglass modes at the free end, which does not exist in fsi elastic gate.
image

image

Xiangyu-Hu:
> # Test on beam with external load
> 
> The geometry and material properties of the beam is the same as in test case 2d_elastic_gate. The force mimicks this time step (0.028s) in fsi, where stress zigzag can be seen near the constraint: image I apply a force that increases linearly at the fixed part and remains constant at the moving part. The maximum force is 0.5. When t < 0.028s, force linearly increases to the target value, then keeps constant thereafter. image
> 
> The beam starts to oscillate under such external pressure force.
> 
> Stress hourglass near the constraint part: image
> 
> Even stronger hourglass modes at the free end, which does not exist in fsi elastic gate. image
> image

@WeiyiVirtonomy  Great! This information is very helpful.

WeiyiVirtonomy:
> > # Test on beam with external load
> > The geometry and material properties of the beam is the same as in test case 2d_elastic_gate. The force mimicks this time step (0.028s) in fsi, where stress zigzag can be seen near the constraint: image I apply a force that increases linearly at the fixed part and remains constant at the moving part. The maximum force is 0.5. When t < 0.028s, force linearly increases to the target value, then keeps constant thereafter. image
> > The beam starts to oscillate under such external pressure force.
> > Stress hourglass near the constraint part: image
> > Even stronger hourglass modes at the free end, which does not exist in fsi elastic gate. image
> > image
> 
> @WeiyiVirtonomy Great! This information is very helpful.

When the beam reaches steady state:

Stress hourglass can still be seen near the constraint:
image

Stronger hourglass near the free end:
image

Xiangyu-Hu:
@WeiyiVirtonomy @DongWuTUM could it the problem come from the constraint. How you constrain the particles? with fix displacement of angle ?

WeiyiVirtonomy:
> @WeiyiVirtonomy @DongWuTUM could it the problem come from the constraint. How you constrain the particles? with fix displacement of angle ?

Fix displacement and rotation

WeiyiVirtonomy:
# Test cases we have so far

## Unit tests for curvature and average curvature calculation
1. test_2d_shell_curvature
2. test_3d_shell_curvature
I'm not sure if you need these cases. I'll keep them for now.

## Basic cases:
1. test_2d_channel_flow_fluid_shell: 2d channel flow
4. test_3d_poiseuille_flow_shell: 3d channel flow

## Verification cases:
1. flow around cylinder shell
2. test_2d_dambreak_elastic_plate_shell
5. test_2d_elastic_gate_fluid_shell
6. fluid_filling_elastic_container
7. hydrostatic_fluid_shell
8. test_3d_dambreak_elastic_plate_shell

## Test cases:
1. flow_around_balloon
2. 3d_bifurcation_pipe
3. flow_around_rigid_plate

DongWuTUM:
@WeiyiVirtonomy Thank you very very much for the details! I will check the `sliding_solid_shell` case. For the result difference between `shell_solid_contact` and `shell_shell_contact`, the constrain may cause this as the constrain is not totally same for the shell and solid.

DongWuTUM:
> @WeiyiVirtonomy @DongWuTUM could it the problem come from the constraint. How you constrain the particles? with fix displacement of angle ?

The hourglass issue is partially from the constraint because there is less hourglass issue when only constraining the displacement. But for the elastic gate case, the result of constraining both displacement and angle is closer to result of volume model.

WeiyiVirtonomy:
Linux-float has failed due to something wrong with get function, but I don't quite understand how this problem came from.

tests/3d_examples/test_3d_poiseuille_flow_shell/poiseuille_flow_shell.cpp
```diff
@@ -259,10 +259,6 @@ int main()
      */
     water_block.addBodyStateForRecording<int>("Indicator");
     water_block.addBodyStateForRecording<Real>("Pressure");
-    shell_boundary.addBodyStateForRecording<double>("Mass");
-    shell_boundary.addBodyStateForRecording<double>("Density");
-    shell_boundary.addBodyStateForRecording<double>("VolumetricMeasure");
-    shell_boundary.addBodyStateForRecording<double>("Thickness");
```
FabienPean-Virtonomy: It failed with float because you make an explicit use of `double`, you need to use `Real`

WeiyiVirtonomy: I see, thanks a lot!

WeiyiVirtonomy:
@DongWuTUM @Xiangyu-Hu As this PR has passed the checks, could we move the contact codes to another PR and merge the FSI part first? Please let me know if there's anything else in fsi that you think needs to be modified

WeiyiVirtonomy:
I have deleted those verification cases in this branch. Please refer to [feature/one_sided_fsi_with_benchmark_examples](https://github.com/Xiangyu-Hu/SPHinXsys/tree/feature/one_sided_fsi_with_benchmark_examples) for the deleted cases.

DongWuTUM:
> @DongWuTUM @Xiangyu-Hu As this PR has passed the checks, could we move the contact codes to another PR and merge the FSI part first? Please let me know if there's anything else in fsi that you think needs to be modified

Hi Weiyi, we could move the contact codes to another PR. And I am checking the FSI part and also the "sliding_solid_shell" case.

WeiyiVirtonomy:
> > @DongWuTUM @Xiangyu-Hu As this PR has passed the checks, could we move the contact codes to another PR and merge the FSI part first? Please let me know if there's anything else in fsi that you think needs to be modified
> 
> Hi Weiyi, we could move the contact codes to another PR. And I am checking the FSI part and also the "sliding_solid_shell" case.

Thanks! I have created a PR for the contact codes: https://github.com/Xiangyu-Hu/SPHinXsys/pull/538

WeiyiVirtonomy:
I thought the hydrostatic fsi failed due to changes in kernel, but actually, it fails randomly due to some other problems. There's already a gtest there. It passed the gtest time in that run of CI test. I haven't found out what caused it to fail yet

DongWuTUM:
> I thought the hydrostatic fsi failed due to changes in kernel, but actually, it fails randomly due to some other problems. There's already a gtest there. It passed the gtest time in that run of CI test. I haven't found out what caused it to fail yet

When I run the hydrostatic fsi on the Windows system, the penetration happens every time. Seems not random? Do you have any idea why this happened, Prof. Hu? @Xiangyu-Hu

WeiyiVirtonomy:
> > I thought the hydrostatic fsi failed due to changes in kernel, but actually, it fails randomly due to some other problems. There's already a gtest there. It passed the gtest time in that run of CI test. I haven't found out what caused it to fail yet
> 
> When I run the hydrostatic fsi on the Windows system, the penetration happens every time. Seems not random? Do you have any idea why this happened, Prof. Hu? @Xiangyu-Hu
> 
> Please check the if gtest does fail the CI when penetration occurs as I also comment in the source code. And remain you add the test case of 3D dambreak with elastic plate.

Could you send me a link of the commit that you are reviewing? I can't find your comments in "files changed".

WeiyiVirtonomy:
> > > > I thought the hydrostatic fsi failed due to changes in kernel, but actually, it fails randomly due to some other problems. There's already a gtest there. It passed the gtest time in that run of CI test. I haven't found out what caused it to fail yet
> > > 
> > > 
> > > When I run the hydrostatic fsi on the Windows system, the penetration happens every time. Seems not random? Do you have any idea why this happened, Prof. Hu? @Xiangyu-Hu
> > > Please check the if gtest does fail the CI when penetration occurs as I also comment in the source code. And remain you add the test case of 3D dambreak with elastic plate.
> > 
> > 
> > Could you send me a link of the commit that you are reviewing? I can't find your comments in "files changed".
> 
> I just commented, didn't change the files.

Sorry, I didn't make it clear. I mean I didn't manage to find your comments on GitHub nor did I receive an notification email for new comments.
I just fixed the Gtest. It should work properly now:

```
The difference between max_disp_analytical and max_disp is nan, which exceeds max_disp_analytical * 15e-2, where
max_disp_analytical evaluates to 7.2666666666666653e-05,
max_disp evaluates to nan, and
max_disp_analytical * 15e-2 evaluates to 1.0899999999999997e-05.
[  FAILED  ] hydrostatic_fsi.dp_2 (961 ms)
[----------] 1 test from hydrostatic_fsi (961 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (961 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] hydrostatic_fsi.dp_2

 1 FAILED TEST
```

The test fails due to diverging to NAN from time to time. I haven't figured out what caused this problem yet.

WeiyiVirtonomy:
Just fixed a bug in the test case hydrostatic_fluid_shell. I think this bug should be the cause of penetration. @DongWuTUM Could you try again and see if it's working now? Thanks

DongWuTUM:
image
image

Hi Weiyi, It is strange that you cannot see the comments. I attached screen shoots; you can also have a look. I will check more, Thanks!

WeiyiVirtonomy:
> image image
> 
> Hi Weiyi, It is strange that you cannot see the comments. I attached screen shoots; you can also have a look. I will check more, Thanks!

No idea why I still can't see those comments, but I pushed some new commits according to the screenshots. As for the kernel, I have tried to create a new kernel with the smoothing length of shell when shell smoothing length > fluid, but it didn't work for the 3d hagen flow test.

DongWuTUM:
> I created a new kernel with the smoothing length of shell when the smoothing length of shell is larger that of fluid, but the result looks weird. Fluid particles are too close to shell particles, and it looks like they are stuck in the gap between shell particles.
> 
> image image
> The velocity/displacement distribution doesn't have much problem though. Shall we discuss about it this week?

Yes, we need to discuss this with Prof. Hu. @Xiangyu-Hu

WeiyiVirtonomy:
> > I created a new kernel with the smoothing length of shell when the smoothing length of shell is larger that of fluid, but the result looks weird. Fluid particles are too close to shell particles, and it looks like they are stuck in the gap between shell particles.
> > image image
> > The velocity/displacement distribution doesn't have much problem though. Shall we discuss about it this week?
> 
> Yes, we need to discuss this with Prof. Hu. @Xiangyu-Hu

Do you have other suggestions except for the kernel selection? I can modify those parts first so that we can merge this branch to master as soon as possible after fixing the kernel issue.

DongWuTUM:
From the coding aspect, there are no issues from my side, and the simulation results are good. Do other reviewers have any concerns or feedback that need to be addressed?

Xiangyu-Hu (approved):

WeiyiVirtonomy:
Why is this PR already merged? I think we still need to discuss the kernel selection

Xiangyu-Hu:
Sorry. I misunderstood Dong's comment. Could we begin a new pull request on the kernel stuff? Xiangyu

PR #466, opened 2023-11-08
Move shell normal update out of relaxation

WeiyiKong321:
A separate class for shell normal update is created since normal direction is not used in stress relaxation


WeiyiKong321:
As far as I see, the only two test cases that use the shell normal direction are _shell_stability_half_sphere_ and _test_thin_structure_dynamics_. Please let me know if there are other tests that need to be modified

Xiangyu-Hu:
You can search the usage by the name of variable. 
Also as the tests cases should have regression test. If all tests passed,  it should be fine.

DongWuTUM (approved):
I have looked through it. It is quite Okay. Thank you, Weiyi.

WeiyiKong321:
> I have looked through it. It is quite Okay. Thank you, Weiyi.

Thanks :)

PR #471, opened 2023-11-13
[SYCL] Enable execution of 3d builds

nR3D:
# Changes
- Added supplementary implementations for 3d methods
- Allocated size of `CellLinkedListKernel::index_head_list_` is now generalized for any dimension, instead of defaulting to the 2d case
- Allocation size of `NeighborhoodDevice` has been increased for 3d builds
- Added test case for 3d dambreak with SYCL execution policy
    - This test case has been written by simply applying the same changes of 2d_dambreak_sycl to 3d_dambreak, nevertheless it has been included to make results more easily reproducible. Alternatively, only one between 2d_dambreak_sycl and 3d_dambreak_sycl could be left as an example for SYCL execution, since they both contain the same changes.

# Benchmark

3d dambreak executed with default parameters and single fp precision:
- Reference TBB execution:  `Total wall time for computation: 37.749652628 seconds.`
- SYCL execution: `Total wall time for computation: 9.485423699 seconds.`


ChiZhangatTUM:
Nice work.

src/shared/common/vector_functions.h
```diff
@@ -131,6 +128,15 @@ inline RealType VecdSquareNorm(const Eigen::Matrix<RealType,Dimension,1>& vec) {
...
+template<class VecType, std::size_t ...Index>
+inline auto VecdFoldingProd_impl(const VecType& vec, std::index_sequence<Index...>) {
+    return ( vec[Index] * ... );
+}
+template<class Type, int Dimension>
+inline Type VecdFoldProduct(const sycl::vec<Type,Dimension>& vec) {
+    return VecdFoldingProd_impl(vec, std::make_index_sequence<Dimension>());
+}
+
 /* SYCL memory transfer utilities */
```
ChiZhangatTUM, line 140: Interesting, you introduce function_impl.

ChiZhangatTUM (approved):

src/shared/common/vector_functions.h
Xiangyu-Hu: What is this for?

nR3D: > - Allocated size of CellLinkedListKernel::index_head_list_ is now generalized for any dimension, instead of defaulting to the 2d case

I used it for this point, essentially `index_head_list_` needs to allocate `all_cells[0] * all_cells[1]` elements for 2d builds and `all_cells[0] * all_cells[1] * all_cells[2]` in the 3d case. I used a [fold expression](https://en.cppreference.com/w/cpp/language/fold) to generalize the product of all elements based on the vector dimension. Having a second function impl is required for expression folding in order to create the index sequence. There are two alternatives to this approach:
1. Using `std::accumulate`, however `sycl::vec` does not provide iterators (`begin` and `end`)
2. Using a simple ternary operator `all_cells[0] * all_cells[1] * (Dimension == 3 : all_cells[2] : 1)`. Which is what `VecdFoldProduct` is computing implicitly, so that there is no need to remember to add the whole expression every time the allocation size of `index_head_list_` is needed

PR #473, opened 2023-11-15
[CI] Build shared libraries

FabienPean-Virtonomy:
Generate shared libraries instead of static one. The objective is to reduce disk size consumed on the GitHub runners by the CI. Currently as static library, the binary code is duplicated many times across all example binaries. Building as shared library should solve that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
       - name: Generate buildsystem using float (No test)
         run: |
           cmake -G Ninja \
+            -D BUILD_SHARED_LIBS=ON\
             -D CMAKE_BUILD_TYPE=Release \
             -D CMAKE_TOOLCHAIN_FILE="${{github.workspace}}/vcpkg/scripts/buildsystems/vcpkg.cmake" \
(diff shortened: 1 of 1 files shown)


Xiangyu-Hu:
The same problem with  fontconfig:x64-osx.

FabienPean-Virtonomy:
Does it reduce the memory consumption ?

ChiZhangatTUM:
Would you please give brief introduction of this PR?

FabienPean-Virtonomy:
It is as per request from @Xiangyu-Hu who needs to have a smaller footprint on the CI disk consumption. A description has been briefly added. I leave it the rest to him.

Xiangyu-Hu:
Something wrong with McOS ?

FabienPean-Virtonomy:
The MacOS GitHub runner image is different between the build job and the test job.
[20230921.1](https://github.com/actions/runner-images/blob/releases/macOS-12/20230921/images/macos/macos-12-Readme.md)
[20231115.2](https://github.com/actions/runner-images/blob/macOS-12/20231115.2/images/macos/macos-12-Readme.md)

The more recent version has Python 3.12, while the older is on 3.11, hence the break. I don't know why there is such discrepancy between runners.

Issue #479, opened 2023-11-27
Update optimization cases before manuscript submission


Xiangyu-Hu:
@Bo-Zhang1995 We need update the optimization cases before manuscript submission.
Also, in the optimization case test_2d_VP_problem1_same_sink_opt, Paramview is not able to read output file 
due to 
</DataArray>
    <DataArray Name="VariationGlobal" type="Float32" Format="ascii">
    -nan(ind) -nan(ind) -nan(ind) -nan(ind) -nan(ind) -nan(ind) -nan(ind)

Xiangyu-Hu:
@Bo-Zhang1995 Please update cases and bug at the assigned branch.

PR #481, opened 2023-11-27
update optimization cases

Bo-Zhang1995:
update the boundary condition, clean the case and renaming the case.
two types of cases are included.
1. the same sink temperature.
2. the heat flux heater.


Xiangyu-Hu:
@Bo-Zhang1995 is the bug shown in issue #479 cleaned?

Bo-Zhang1995:
yeah，i checked the result in paraview that shows normal. Google test could also pass.On 28 Nov 2023, at 07:12, Xiangyu Hu ***@***.***> wrote:﻿
@Bo-Zhang1995 is the bug shown in issue #479 cleaned?

Bo-Zhang1995:
> @Bo-Zhang1995 is the bug shown in issue #479 cleaned?

yeah，i checked the result in paraview that shows normal. Google test could also pass.

PR #482, opened 2023-11-28
use force and force_prior to replace acc and acc_prior

ZhentongWang:


Xiangyu-Hu:
@ZhentongWang so the time integration methods are updated correspondingly?

ZhentongWang:
> @ZhentongWang so the time integration methods are updated correspondingly?

the time integration you mentioned is like the viscous force in the viscous force calculation, velocity update in the pressure relaxation and density update in the density relaxation and so on. If so, I have changed and used force and force_prior to update all variables in fluid and solid dynamics. 

For example, the velocity updated as follows
vel_[index_i] += (force_prior_[index_i] + force_[index_i]) / mass_[index_i] * dt;

Xiangyu-Hu (approved):

PR #484, opened 2023-12-01
Plastic hourglass control

DongWuTUM:
Generalized essentially non-hourglass control formulation based on volumetric-deviatoric stress decomposition.


Xiangyu-Hu:
Some thing wrong for Mac and windows systems.

Xiangyu-Hu:
@DongWuTUM As this is also a merge to core library, I suggest that the new source and test cases be moved to the core library and tests.

Xiangyu-Hu (approved):

PR #475, opened 2023-11-18
revise the continuum dynamics

Shuaihao-Zhang:
I revised the continuum dynamics, mainly the following three parts,
(1) modified the time integration to obey the position based Verlet scheme.
(2) modified the Riemann solver and only applied numerical dissipation on the pressure part.
(3) the pressure in the Riemann solver is calculated by the trace of stress tensor rather than using the equation of state, because the plastic deformation will also affect the hydrostatic pressure, while the equation of state only consider the elastic part.
The accuracy of the revised version is improved. For example, when modelling the 3D column collapse to study the relationship of run-out distance and the aspect of the column:
image
The results obtained by the previous version is:
image
The results obtained by the current version is:
image

PR #486, opened 2023-12-04
revise the continuum dynamics -- solve the conflicts 

Shuaihao-Zhang:
Since some new PRs are merged, there are some conflicts.

I solved the conflicts based on the latest master version. I also simplified the code (delete many unnecessary code blocks and comments) and improved the code quality.

Most of the comments from #475  are considered in this version, except the following two points:

(1) I still use three step relaxations. I try your comments from #475. Combining any of the two processes will change the order of updating the variables and the affect the results.

As you can see here, we have three different particle dynamics (the interaction in each step), and some variables are updated between each two particle dynamics. I tried and found the current sequence gives the best result compared with experimental data and other people's numerical results.
image

(2) I still keep the template class for Riemann solver. Although I just need acoustic Riemann solver for the uploaded two cases, I may need different Riemann solvers in the following studies.


Shuaihao-Zhang:
> I am wondering why there is no way to put the computation of velocity gradient behind density change rate. It seems quite easy by combining computing velocity gradient and density change together and update strain by two steps as the density, and then update stress for momentum equation. It is the same as other solid dynamics updating and should be equivalent to the present version.

Yes, you are right. 
I have modified the code, and now we just have two steps of relaxation process. 
Thanks for your comments.

Xiangyu-Hu:
@Shuaihao-Zhang Thanks for the revision. Before doing detailed review,  I am wondering whether there are classes or methods are not used by the tests. If there is something like that, you can either build up a test for it or simply delete it. Not tested code pieces are no use but big troubles for maintaining.

Shuaihao-Zhang:
> @Shuaihao-Zhang Thanks for the revision. Before doing detailed review, I am wondering whether there are classes or methods are not used by the tests. If there is something like that, you can either build up a test for it or simply delete it. Not tested code pieces are no use but big troubles for maintaining.

@Xiangyu-Hu  Thanks for your suggestion. I have deleted the codes that are not used by test cases.

PR #502, opened 2024-01-02
Change mass_ to "Mass" instead of "MassiveMeasure"

WeiyiVirtonomy:
Particle property `mass_` is changed to the volumetric mass, instead of area mass, so that `force_` and `force_prior_` now represent the volumetric force.

### DIFF
--- a/src/for_3D_build/particles/beam_particles.h
+++ b/src/for_3D_build/particles/beam_particles.h
@@ -29,13 +29,13 @@
 #ifndef BEAM_PARTICLES_H
 #define BEAM_PARTICLES_H
 
+#include "base_body.h"
 #include "base_particles.h"
 #include "base_particles.hpp"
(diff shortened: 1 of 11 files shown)


WeiyiKong321:
The schemes of some source files are changed by the auto scheme correction, I can change them back if necessary

WeiyiVirtonomy:
> I have checked. It is Okay for me.

`DampingPairwiseInner` uses `Vol_[index_i] / mass_[index_i]`, so for shell I have to change all the physical damping used to physical damping * thickness. Do you think it would be better to use `ParticleVolume(index_i) ` in `DampingPairwiseInner` instead?

Xiangyu-Hu:
@WeiyiVirtonomy You can make real pull request for  the changes.

Xiangyu-Hu (approved):

PR #503, opened 2024-01-09
[SYCL] Use Eigen classes within SYCL kernels

nR3D:
So far SYCL vector types (`sycl::vec`) have been used as a replacement for Eigen matrices and arrays, since the latter are directly supported for SYCL kernel execution. This led to the introduction of a parallel set of data types (e.g. `DeviceVecd` instead of `Vecd`) that could be used within SYCL kernels.
However, Eigen provides many algebraic methods that SYCL vectors lack, additionally the support for two different vector types would mean a less maintainable codebase. Ideally, using Eigen for both CPU and GPU execution is the option to pursue (as long as the performance gap w.r.t. SYCL vectors is not excessive).
The following changes  enable an initial support for Eigen data types within SYCL kernels, however many aspect are still to be considered, so I will keep this pull request as a draft for the moment being:
- [X] `test_2d_dambreak` is working as expected (with ca. 4% slowdown compared to SYCL vectors)
- [x] `test_3d_dambreak` is compiling correctly and it executes for the first second of simulation time, however it then segfaults by accessing a wrong GPU address. The simulation results are correct up until that point, and the same simulation runs fine with SYCL vectors.
    - Solved by increasing the allocation size of NeighborhoodDevice
- [X] When trying to compile using the cmake option "Debug" the compiler fails to generate PTX files (in the case of nvidia ahead-of-time compilation), which makes it more difficult to debug the 3d case, however "Release" compilation works fine. This could be the result of some code that Eigen uses internally which is considered undefined behavior to compile for device execution, and/or it's not supported by the compiler.
    - Solved by defining SYCL_DEVICE_ONLY or by switching to the git version of Eigen and using EIGEN_USE_SYCL
- [X] Eigen uses internally generic pointers for some functions (like `min`/`max`) which are not supported by SYCL (https://support.codeplay.com/t/eigen-support-only-tensor-module-or-arrays-and-matrices-as-well/561/5) so those methods have to be replaced by another implementation. However, re-implementing Eigen methods is against the objective of this PR. Ultimately, a work-around to make all Eigen methods work is needed.
    - Solved by defining SYCL_DEVICE_ONLY or by switching to the git version of Eigen and using EIGEN_USE_SYCL

https://gitlab.com/libeigen/eigen/-/issues/2763
This is a recently-opened discussion with additional resources and insights on SYCL support by Eigen.


Xiangyu-Hu:
@nR3D we can check the macro in eigen for using https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/Core#L111

Xiangyu-Hu:
@nR3D another place for sycl is https://eigen.tuxfamily.org/dox-devel/Macros_8h_source.html

src/shared/common/base_data_type.h
```diff
@@ -39,6 +39,12 @@
 #include <map>
 #include <vector>
 
+// Avoid automatic memory allocations
```
Xiangyu-Hu, line 42: try to use SYCL_DEVICE_ONLY
as referred from https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/util/Macros.h#L851

nR3D: Thank you, this seems to solve the usage of generic pointers and debug compilation. However I had to switch to the git version of Eigen, since their latest published release (3.4.0) does not have those macros. I will later try to debug the 3d case to see if I can find the point in which it dereferences an illegal address.

Xiangyu-Hu: This macro is also in the current version of eigen,
please see it in /vcpkg/installed/x64-linux/include/eigen3/Eigen/src/Core/util/Macros.h

nR3D: The problem with 3.4.0 seems to be that enabling EIGEN_USE_SYCL throws errors at compile-time, which have been fixed in their master branch, however defining SYCL_DEVICE_ONLY works also for 3.4.0

https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/src/Core/util/Macros.h#L568
Judging from this definition, it seems that SYCL_DEVICE_ONLY is a macro defined by Eigen when EIGEN_USE_SYCL is enabled and device code is being compiled.

We could either:
- define SYCL_DEVICE_ONLY with 3.4.0, until a new version is released and then we can switch to EIGEN_USE_SYCL
- define EIGEN_USE_SYCL and switch Eigen's dependency to its git version until a new release is available

However, defining SYCL_DEVICE_ONLY without EIGEN_USE_SYCL will exclude lines of codes tailored for SYCL (e.g. https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/Core#L238).

Xiangyu-Hu: If the interface for the codes tailored for SYCL is the same for those with SYCL_DEVICE_ONLY, I would not mind  the loss of some performance.

nR3D: - https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/Core#L238
- https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/Core#L116

These are some of the parts that would be excluded by not using EIGEN_USE_SYCL, so the entirety of [src/Core/arch/SYCL](https://gitlab.com/libeigen/eigen/-/tree/3.4.0/Eigen/src/Core/arch/SYCL), and some library inclusions, most importantly <CL/sycl.hpp> which we would need to include separately (even though we already include <sycl/sycl.hpp> but Eigen expects the CL namespace).

[src/Core/MathFunctions.h](https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/src/Core/MathFunctions.h) and [src/Core/functors/UnaryFunctors.h](https://gitlab.com/libeigen/eigen/-/blob/3.4.0/Eigen/src/Core/functors/UnaryFunctors.h) are using SYCL_DEVICE_ONLY and that seems to be enough to make it run.

For what concerns performance, the difference seems negligible, at least for those functions used by dambreak.

Xiangyu-Hu: @nR3D so we use SYCL_DEVICE_ONLY now and change it to EIGEN_USE_SYCL after Eigen updated for sycl namespace?

nR3D:
I have solved the issue in the 3d case, it turned out to be a problem with the allocation size of the neighborhood list, which was too small. As discussed in previous PRs, the solution would be to define a method that determines the maximum number of particles contained withing the kernel's smoothing radius, or by removing the usage of a neighborhood list all together.

At this point all the issues with Eigen seem to have been solved, so I will remove the draft status from the PR.

Xiangyu-Hu:
@nR3D Great to hear it.

PR #524, opened 2024-02-05
Fix/more sortable variables

Xiangyu-Hu:
Try to avoid problems due to particle sorting.


Xiangyu-Hu:
something is wring on the modified files.

PR #529, opened 2024-02-16
[SYCL] Substitute system configuration with direct search

nR3D:
# Changes

SYCL code will now iterate over the entire list of particles contained within neighboring cells instead of relying on pre-computed kernel values coming from a configuration update step.

A new method `forEachNeighbor` has been added to `CellLinkedListKernel`, which replaces `searchNeighborsByParticles` and the need to update configuration variables.
In fact, all variables and methods concerning configuration information for device code have been removed.

# Benchmarks

A performance comparison between the old configuration-based execution and the new direct search version shows an improvement up to 6x for larger simulations.

Benchmark

More importantly, removing configuration variables significantly reduced GPU memory usage. For instance, 2 million particles required up to 12GiB of RAM on the old version, while it now only takes 900MiB, ultimately enabling larger simulation on the same graphic cards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,9 @@ add_subdirectory(modules)
     add_subdirectory(tests)
 endif()]]
 add_subdirectory(tests/2d_examples/test_2d_dambreak)
+add_subdirectory(tests/3d_examples/test_3d_dambreak)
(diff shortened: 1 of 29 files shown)


Xiangyu-Hu:
Great! It is great to hear that direct search is actually much faster.

PR #535, opened 2024-02-24
Fix/limiting transport velocity

Xiangyu-Hu:
transport velocity formulation can be limited according to zeroth inconsistency error.


Xiangyu-Hu:
@Lisu-lisa @Bo-Zhang1995 @ZhentongWang could you have a look the pull request and approve it if there is no issue? Thanks.

Lisu-lisa (approved):

PR #537, opened 2024-02-28
Refactory/check oldroyd b

Xiangyu-Hu:
I have found a bug in the throat case, in which the dissipation term has been wrongly added in contact interaction.


Xiangyu-Hu:
@TheodorH999 You can use the new class for computing velocity gradient in your project if you like.
Certainly You can still use you own for now.

PR #551, opened 2024-03-30
Refactory/revise data assemble

Xiangyu-Hu:
Try to used data assemble tuple explicitly only once so that we can easily add more data types.
This is not done for mesh data yet as they need new formulation on variable containers.


Xiangyu-Hu:
@Lisu-lisa Could have a check and approve the pull request if there is no issue? Thanks.

Lisu-lisa (approved):

PR #572, opened 2024-05-06
Add regression data for fluid-shell dambreak

WeiyiVirtonomy:
Generate regression data for the test case dambreak_elastic_plate_shell.


WeiyiVirtonomy:
@DongWuTUM Hi, Dong, could you check the regression data of the test case dambreak_elastic_plate_shell when you have time? I   am not so familiar with the regression tools, please let me know if there is something needs to be fixed.

DongWuTUM:
@WeiyiVirtonomy No problem. I will have a look. Seems Okay. If there is a problem, I will fix it.

DongWuTUM:
@WeiyiVirtonomy Hi, Weiyi. The regeression data you uploaded are Okay, through I regenerate again. Only several data need to be pre-strored in source code. 

In case you need to use regression tool again: 1. You can learn form the other test cases, and choose a apporate regression method; 2. If the regression cannot converge, you can use the larger threshold at "generateDataBase(1.0e-3)"; 3. Some details you can find in https://www.sphinxsys.org/html/regression_test.html.

DongWuTUM (approved):

PR #575, opened 2024-05-27
fix a bug for particle reload in debug mode

Xiangyu-Hu:
The assert which is checked in debug mode is wrong.
Only changes are on L100 and L120.


Xiangyu-Hu:
@FengWang3119 @ChenxiZhaoTUM  The bug suggest why debug mode crash when using reload.

FengWang3119 (approved):

PR #596, opened 2024-06-15
modify T shaple flow case

ShuoguoZhangTUM:
impose pressure boundary condition and bidirectional buffer


DrChiZhang (approved):

Xiangyu-Hu:
@DrChiZhang  @ShuoguoZhangTUM  this pull request should be reverted as it even not able to compile.

PR #597, opened 2024-06-16
Add variable and derived variable to write

Xiangyu-Hu:
1. real_bodies vector in sph system is protected.
2. move the variable and derived variable add to write into the body state recording for move expressive functions.


Xiangyu-Hu:
I will reopen later after a stage is finished.

Lisu-lisa (approved):

PR #602, opened 2024-06-22
revise the registeration of dicrete variables.

Xiangyu-Hu:
1. I first make all discrete variables shared and let the data field directly saved in the variable. 
2. I split the geometric data for particle generation and its final form. That is, the geometric data created using dynamic memory will be copied to static memory after all particles generated.


Xiangyu-Hu:
I will reopen after a stage is done.

Xiangyu-Hu:
I close this branch as it seems polluted by so many changes. I should stick on to incremental changes.

PR #604, opened 2024-06-27
Shell code simplification

DongWuTUM:
Add the global F, F_bending, and transformation matrix from the initial local coordinates to the current local coordinates to simplify the calculation.


Xiangyu-Hu:
@DongWuTUM could you have a check on speed up?

DongWuTUM:
> @DongWuTUM could you have a check on speed up?

I tested the 3D arch. Total wall time for computation: Original 33.53 s - Now 31.12 s, speed-up 7.2%. 

Arch is set to be simulated without hourglass control in the test case. The added global F and F_bending can improve the efficiency of the hourglass algorithm. Thus, I also tested the 3D arch **with** the hourglass control. 
Total wall time for computation: Original 63.61 s - Now 44.84 s, speed-up 34.2%.

Xiangyu-Hu (approved):

PR #616, opened 2024-07-09
add the 2d LDC case with RKGC correction.

Bo-Zhang1995:
Here I add the 2d LDC case with RKGC correction, representing the internal flow.


Bo-Zhang1995:
> @Bo-Zhang1995 It seems the new case is not able to run.
I am modifying it now.

Xiangyu-Hu (approved):

PR #617, opened 2024-07-09
aligned box with defined upper bound

Xiangyu-Hu:
We define the upper bound axis for aligned box when the latter is constructed. So that we do not need specify it later in constructing particle dynamics.


Xiangyu-Hu:
@ChenxiZhaoTUM @ShuoguoZhangTUM Please have a check to the modifications relevant  to your recent works.

ShuoguoZhangTUM (approved):

PR #619, opened 2024-07-12
Non-dimensionalised contact density

WeiyiVirtonomy:
1. Replace mass in the calculation of contact densities by volume
2. Change contact stiffness to `rho0 * c0 * c0`
3. For the contact between two elastic bodies, the contact stiffness is averaged by `2 * K1 * K2 / (K1 + K2)`


WeiyiVirtonomy:
@DongWuTUM @Xiangyu-Hu 
Currently, I'm multiplying the contact stiffness `c0 * c0` by reference density in contact force classes. Would you prefer to change the definition in `base_material`?

DongWuTUM:
> @DongWuTUM @Xiangyu-Hu Currently, I'm multiplying the contact stiffness `c0 * c0` by reference density in contact force classes. Would you prefer to change the definition in `base_material`?

Yes, it should be changed accordingly. And the name 'repulsion_density_' should be changed. How about 'repulsion_factor_'? @Xiangyu-Hu @WeiyiVirtonomy

WeiyiVirtonomy:
> 'repulsion_factor_'?

'repulsion_factor_' sounds good for me

WeiyiVirtonomy:
Renaming is finished
I only changed the variable names, the class names still remained `ContactDensitySummation`

Winterless11:
> Renaming is finished I only changed the variable names, the class names still remained `ContactDensitySummation`

Sorry for asking changing again. Could you also change the class name and the file name 'repulsion_density_summation'?

WeiyiVirtonomy:
> > Renaming is finished I only changed the variable names, the class names still remained `ContactDensitySummation`
> 
> Sorry for asking changing again. Could you also change the class name and the file name 'repulsion_density_summation'?

Sure, will do that

WeiyiVirtonomy:
Something that might need to discuss:
When I was testing on the contact of a stiff and a soft body with a Young's modulus ratio of 1e4 (which is typical for devices and tissues), when the impact velocity is higher than the sound speed of the soft body, the averaged contact stiffness is not enough to prevent penetration.
I'll upload the case before Friday so that we can have a discussion.

Xiangyu-Hu:
> Something that might need to discuss: When I was testing on the contact of a stiff and a soft body with a Young's modulus ratio of 1e4 (which is typical for devices and tissues), when the impact velocity is higher than the sound speed of the soft body, the averaged contact stiffness is not enough to prevent penetration. I'll upload the case before Friday so that we can have a discussion.

It seems that we are not able to use the exact Youngs modules for the device, 
other than a weakly compressible version of it. 

To match the soft-stiff relation, a possible approach is: 
begin from the soft tissue with its Youngs modules, then propose an affordable Poisson ratio, near 0.5, to determine the bulk modules, then using the same bulk modules for the stiff device. One can obtain the artificial Youngs modules (still much larger (about 10 times?) than that of the tissue) of the device with the real Poisson ratio (around 0.3).

In this way, the same time step can be used for both device and tissue. And, one expect that the contact between them leads to very small deformation for the device and fairly large one for the tissue.

WeiyiVirtonomy:
# A test case of stiff-soft contact
The half sphere is made of soft material, while the brick is made of stiff material. A constant velocity is applied to the right end of the brick to push it moving toward the sphere.
`test_3d_stiff_soft_contact`

E_stiff/E_soft = 1e4

1. Velocity / c_soft = 2
The contact force cannot prevent penetration 
Screenshot 2024-07-25 112219

2. Velocity / c_soft = 1
The shell is pushed backward due to the contact force, but the shell is very unstable
Screenshot 2024-07-25 112520

3. Velocity / c_soft = 0.5
Seems good
Screenshot 2024-07-25 112329

4. Same bulk modulus, E_stiff / E_soft = 20, Velocity / c = 1
The stiff material cannot respond to the velocity applied to the right end promptly
Screenshot 2024-07-25 112626

4. Same bulk modulus, E_stiff / E_soft = 20, Velocity / c = 0.2
Seems good
Screenshot 2024-07-25 112738

WeiyiVirtonomy:
> > Something that might need to discuss: When I was testing on the contact of a stiff and a soft body with a Young's modulus ratio of 1e4 (which is typical for devices and tissues), when the impact velocity is higher than the sound speed of the soft body, the averaged contact stiffness is not enough to prevent penetration. I'll upload the case before Friday so that we can have a discussion.
> 
> It seems that we are not able to use the exact Youngs modules for the device, other than a weakly compressible version of it.
> 
> To match the soft-stiff relation, a possible approach is: begin from the soft tissue with its Youngs modules, then propose an affordable Poisson ratio, near 0.5, to determine the bulk modules, then using the same bulk modules for the stiff device. One can obtain the artificial Youngs modules (still much larger (about 10 times?) than that of the tissue) of the device with the real Poisson ratio (around 0.3).
> 
> In this way, the same time step can be used for both device and tissue. And, one expect that the contact between them leads to very small deformation for the device and fairly large one for the tissue.

I tried to use a pseudo-Young's modulus for the stiffer material, which is about 20 times the soft ones, but the characteristic velocity still needs to be smaller than the sound speed, otherwise, the information cannot propagate to the other parts of the solid in time.
I'm not sure if the pseudo modulus is appropriate for customer projects.

Xiangyu-Hu:
> > > Something that might need to discuss: When I was testing on the contact of a stiff and a soft body with a Young's modulus ratio of 1e4 (which is typical for devices and tissues), when the impact velocity is higher than the sound speed of the soft body, the averaged contact stiffness is not enough to prevent penetration. I'll upload the case before Friday so that we can have a discussion.
> > 
> > 
> > It seems that we are not able to use the exact Youngs modules for the device, other than a weakly compressible version of it.
> > To match the soft-stiff relation, a possible approach is: begin from the soft tissue with its Youngs modules, then propose an affordable Poisson ratio, near 0.5, to determine the bulk modules, then using the same bulk modules for the stiff device. One can obtain the artificial Youngs modules (still much larger (about 10 times?) than that of the tissue) of the device with the real Poisson ratio (around 0.3).
> > In this way, the same time step can be used for both device and tissue. And, one expect that the contact between them leads to very small deformation for the device and fairly large one for the tissue.
> 
> I tried to use a pseudo-Young's modulus for the stiffer material, which is about 20 times the soft ones, but the characteristic velocity still needs to be smaller than the sound speed, otherwise, the information cannot propagate to the other parts of the solid in time. I'm not sure if the pseudo modulus is appropriate for customer projects.

Good. The same as the weakly compressible fluid model, v_max should be around 1/10 sound speed. 0.2 is still a bit too large.

Xiangyu-Hu:
> 1. Replace mass in the calculation of contact densities by volume
> 2. Change contact stiffness to `rho0 * c0 * c0`
> 3. For the contact between two elastic bodies, the contact stiffness is averaged by `2 * K1 * K2 / (K1 + K2)`

@WeiyiVirtonomy you need make the pull request ready for review not just draft.

Xiangyu-Hu (approved):

PR #624, opened 2024-07-15
Fix the bug for level set generation in debug mode

Xiangyu-Hu:
@FengWang3119 @ChenxiZhaoTUM Please have look if this bug fix solve your problem on level set generation.

### DIFF
--- a/src/for_3D_build/meshes/mesh_with_data_packages.hpp
+++ b/src/for_3D_build/meshes/mesh_with_data_packages.hpp
@@ -140,7 +140,7 @@ void MeshWithGridDataPackages<PKG_SIZE>::
         for (int j = 0; j != pkg_size; ++j)
             for (int k = 0; k != pkg_size; ++k)
             {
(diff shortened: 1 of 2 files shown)


FengWang3119:
Still not working for my case.
It is strange, for DL=120 DL=60 it is OK for water_block.defineBodyLevelSetShape(), but for DL=100 it is keeping looping.
QQ_1721067700460
Is is because we use the EXTRUDE?

Xiangyu-Hu:
> Still not working for my case. It is strange, for DL=120 DL=60 it is OK for water_block.defineBodyLevelSetShape(), but for DL=100 it is keeping looping. QQ_1721067700460 Is is because we use the EXTRUDE?

I guess only a close look can find the reason.

Xiangyu-Hu:
Or it is a different issue?

Xiangyu-Hu:
Anyway, I found the previous bug on debug mode solved.

Lisu-lisa (approved):

PR #634, opened 2024-08-08
[Test] solid multi-resolution

WeiyiVirtonomy:
Aim: test the behavior of solid multi-resolution simulation

- [x] Use minimum smoothing length in time step size
- [ ] Use smoothing_length_ / h_ratio in `Integration1stHalf` (maybe not necessary)
- [ ] Modify random damping

Test case 1: 2d beam subjected to a uniform gravity
Left: multi-resolution, right: single-resolution
image

Tip position:
image


WeiyiVirtonomy:
The `SplittingCellLists` is not so clear to me, can we have a discussion this Friday?

Xiangyu-Hu:
> The `SplittingCellLists` is not so clear to me, can we have a discussion this Friday?

Sure.

WeiyiVirtonomy:
Added a temporary split cell list to multi-resolution:
image

WeiyiVirtonomy:
## Question about inner relation for damping
From my understanding, the main problem we have in the splitting algorithm for multi-resolution is that: 
Consider two particles with dp1 < dp0 and a distance of two cells. Though particles of the same mesh level won't exist in their neighborhood at the same time, particles with dp0 can be present in both of their neighborhoods, hence leading to memory conflict for neighbors with dp0.

To solve this problem, we build a special inner neighbor for the splitting method. A particle i can only see neighbors with mesh_level_j >= mesh_level_i. Since this particle pair is only executed once, neighbors in the same level need to be consistent with it. Hence, for mesh_level_j = mesh_level_i, the particle i can only see particles j with index_j > index_i. Each particle pair are only executed once.

My question about this is that, in the single-resolution damping, each particle pair is executed twice, i.e. for a particle pair (i, j), vel_i and vel_j are updated once when executing damping for particle i and neighbor j, and updated again when executing damping for particle j and neighbor i. How can we keep consistent with the single-resolution damping?

## Refinement level 2 fails
I set the refinement level as two and tried to run the simulation without damping. However, it failed soon after the simulation starts:
image

It was fine with refinement level 1. I haven't figured out what caused the failure yet

@DongWuTUM Shall we have a discussion some time this week?

DongWuTUM:
@WeiyiVirtonomy Yes, we can have a discussion this afternoon (11th of Sep.) at 14:30, if it also works for you.

WeiyiVirtonomy:
> @WeiyiVirtonomy Yes, we can have a discussion this afternoon (11th of Sep.) at 14:30, if it also works for you.

Great, see you this afternoon

WeiyiVirtonomy:
## Refinement level 2

The failure was caused by the pairwise damping in `Integration1stHalf`. After changing the reference smoothing length to `smoothing_length_ / h_ratio_[index_i]`, the simulation can go on, but there are still singularities in the refined part.

image

I think we need to consider what W_ij to use for AdaptiveInner. We are now using the kernel of the particle itself:

`Real weight = distance < kernel_->CutOffRadius(i_h_ratio) ? kernel_->W(i_h_ratio, distance, displacement) : 0.0;`

Maybe for `Integration1stHalf` it needs to be

`Real weight = kernel_->W(h_ratio_min, distance, displacement);`

WeiyiVirtonomy:
# Test with refinement level 2
Since pairwise damping stress cannot be directly applied to multi-resolution bodies, I used the right Cauchy damping.
Without using physical damping:
image

Using physical damping:
image

To match the damping algorithm of single-resolution where a pair of particles is damped twice, in this test the damping for multi-resolution is executed twice.

The result seems pretty close.

WeiyiVirtonomy:
I tried to compare single-multi-resolution behavior by the 3D cantilever beam example used in the damping paper, but the 3D multi-resolution particle distribution does not look so good after relaxation. It seems that there's a gap in the interface of the refinement region
image

@DongWuTUM Do you have any idea about this problem?

WeiyiVirtonomy:
# Results
I tried to use the 3d cantilever beam as a validation case. The reference data is from [damping](https://doi.org/10.1016/j.jcp.2022.111105).
image

## Convergence study
The deflection of the middle point at the right end is used to study the spatial convergence. To compare with the data from reference paper (Figure 3), the curve is shifted to match $y_0=0.02$.

- Uniform resolution
image

- Multi-resolution with refinement level 2
image

From the curve, we may conclude that the result is nearly converged with $dp=t/32$ and $dp=t/14~t/56$ for uniform and multi-resolution respectively. 

## No damping
The deflection curves of single resolution with $dp=t/40$, multi-resolution with $dp=t/16~t/64$, and reference paper (Figure 3) are plotted in the figure below. The curves have deviations in later periods, but the frequency of mr and uniform resolution is quite close. 
image

## Damping
The physical coefficient is taken as $\eta_0$ in the paper. The deflection curves of single resolution with $dp=t/32$ and multi-resolution with $dp=t/14~t/56$ are shown in the figure below.
Since each particle-pair is can only see each other once in the new inner configuration for multi-resolution, I executed damping twice for the multi-resolution case. However, the convergence speed of the mr case is still much slower than the uniform resolution.
image
I'll check what caused the difference. @DongWuTUM

DongWuTUM:
> # Results
> I tried to use the 3d cantilever beam as a validation case. The reference data is from [damping](https://doi.org/10.1016/j.jcp.2022.111105). image
> 
> ## Convergence study
> The deflection of the middle point at the right end is used to study the spatial convergence. To compare with the data from reference paper (Figure 3), the curve is shifted to match y 0 = 0.02 .
> 
> * Uniform resolution
> 
> image
> * Multi-resolution with refinement level 2
> 
> image
> From the curve, we may conclude that the result is nearly converged with d p = t / 32 and d p = t / 14   t / 56 for uniform and multi-resolution respectively.
> 
> ## No damping
> The deflection curves of single resolution with d p = t / 40 , multi-resolution with d p = t / 16   t / 64 , and reference paper (Figure 3) are plotted in the figure below. The curves have deviations in later periods, but the frequency of mr and uniform resolution is quite close. image
> 
> ## Damping
> The physical coefficient is taken as η 0 in the paper. The deflection curves of single resolution with d p = t / 32 and multi-resolution with d p = t / 14   t / 56 are shown in the figure below. Since each particle-pair is can only see each other once in the new inner configuration for multi-resolution, I executed damping twice for the multi-resolution case. However, the convergence speed of the mr case is still much slower than the uniform resolution. image I'll check what caused the difference. @DongWuTUM

Thank you. The deflection curves without damping are quit close! Multi-resolution is slower to reach quasi-static than single-resolution for this 3D case, but I see multi-resolution is faster for 2D case. You can have a try to use larger physical coefficient η0 to reach quasi-static. 

Not long ago you said "the 3D multi-resolution particle distribution does not look so good after relaxation." How did you solve it?

WeiyiVirtonomy:
> > # Results
> > I tried to use the 3d cantilever beam as a validation case. The reference data is from [damping](https://doi.org/10.1016/j.jcp.2022.111105). image
> > ## Convergence study
> > The deflection of the middle point at the right end is used to study the spatial convergence. To compare with the data from reference paper (Figure 3), the curve is shifted to match y 0 = 0.02 .
> > 
> > * Uniform resolution
> > 
> > image
> > * Multi-resolution with refinement level 2
> > image
> > From the curve, we may conclude that the result is nearly converged with d p = t / 32 and d p = t / 14   t / 56 for uniform and multi-resolution respectively.
> > ## No damping
> > The deflection curves of single resolution with d p = t / 40 , multi-resolution with d p = t / 16   t / 64 , and reference paper (Figure 3) are plotted in the figure below. The curves have deviations in later periods, but the frequency of mr and uniform resolution is quite close. image
> > ## Damping
> > The physical coefficient is taken as η 0 in the paper. The deflection curves of single resolution with d p = t / 32 and multi-resolution with d p = t / 14   t / 56 are shown in the figure below. Since each particle-pair is can only see each other once in the new inner configuration for multi-resolution, I executed damping twice for the multi-resolution case. However, the convergence speed of the mr case is still much slower than the uniform resolution. image I'll check what caused the difference. @DongWuTUM
> 
> Thank you. The deflection curves without damping are quit close! Multi-resolution is slower to reach quasi-static than single-resolution for this 3D case, but I see multi-resolution is faster for 2D case. You can have a try to use larger physical coefficient η0 to reach quasi-static.
> 
> Not long ago you said "the 3D multi-resolution particle distribution does not look so good after relaxation." How did you solve it?

That 2d beam was too thin, so it's natural that it took more time to reach the static state.
I think it would work with a larger $\eta$, but the difference between the new damping method and the old one is larger than I expected. I'll check the code again later
The relaxation problem was caused by the use of `RelaxationStepLevelSetCorrectionInner`, I replaced it with `RelaxationStepInner`. (See this issue: #663)

WeiyiVirtonomy:
The sad news is that the computational time of multi-resolution damping is much longer than that of single-resolution.

Single resolution of t/16, 1 mesh level: 23s without damping, 32s with damping
Refinement level 1 of t/8 ~ t/16, 1 mesh level: 13s without damping, 77s with damping
Refinement level 2 of t/4~t/16, 2 mesh levels: 13s without damping, 97s with damping

The gap possibly comes from the cell linked list. The cell-linked list of the single resolution t/16 looks like this:
image

While the cell-linked list of the refinement level 1 t/8 ~ t/16 looks like this:
image

The single resolution has better parallelization.
When the refinement level is higher than 1, there will be more than one mesh level and the condition becomes worse because the splitting algorithm of different mesh levels needs to be run in sequence to avoid conflict. For the single-resolution there are 27 split cells , while for mr with refinement level 2 there are 57 split cells.

Xiangyu-Hu:
@WeiyiVirtonomy No. It is this paper. http://dx.doi.org/10.1016/j.jcp.2013.12.061

Xiangyu-Hu:
and this https://link.springer.com/article/10.1007/s00450-011-0167-z

Xiangyu-Hu:
Actually, they all are connected.

WeiyiVirtonomy:
> and this https://link.springer.com/article/10.1007/s00450-011-0167-z in section 5

Thank you so much!

WeiyiVirtonomy:
@Xiangyu-Hu Hello, Prof. Hu, sorry for disturbing you again. I've started to read the papers about the adaptive method, but it is still unclear in my mind how these data structures and time stepping are related to the damping algorithm. It would be very helpful if you could briefly summarize the ideas you proposed last week so that I could have a better understanding of the problem before Friday's meeting. Thank you so much

Xiangyu-Hu:
> @Xiangyu-Hu Hello, Prof. Hu, sorry for disturbing you again. I've started to read the papers about the adaptive method, but it is still unclear in my mind how these data structures and time stepping are related to the damping algorithm. It would be very helpful if you could briefly summarize the ideas you proposed last week so that I could have a better understanding of the problem before Friday's meeting. Thank you so much

We can come tomorrow after or next week, as we will have no discussion on Friday due to another meeting.

WeiyiVirtonomy:
## Further check on the cause of high computational time

I compared the results with a single resolution $dp = h/16$ and a multi-resolution ${dp}_0 = h/8$ and ${dp}_1=h/16$. The whole domain is the refinement region so that all the particles in the multi-resolution case are of ${dp}_1=h/16$. Hence, the particle number and inner configurations of the mr case is nearly the same as the single resolution one.

image
image

Left: initial particle distribution of the single resolution case
Right: initial particle distribution of the multi-resolution case

For the multi-resolution case, the cell level is set as:

1. Only level 0 with (17, 8, 8) cells
2. Level 0 with (17, 8, 8) and level 1 with (29 11 11).

For the second type, all the particles belong to level 1.

The cell number of the single-resolution case is (29 11 11).

## Computation time
The particle number, inner neighbor number and computational time of the three cases are:

| Type | Particle number | Average inner neighbor | Average damping inner neighbor | Computational time | Damping time |
|-------|------------------|----------------------------|--------------------------------------|-----------------------|-----------------|
| Single | 14336 | 44 | 22 | 37 | 7 |
| MR, one level | 14326 | 43 | 21 | 63 | 29 |
| MR, two level | 14326 | 43 | 21 | 39 | 8 |

The damping time of the two level one is very close to the single-resolution one, so we might conclude that one cause of poor efficiency is the poor cell-parallelization when we only use one cell level.

##  Weird damping behavior
The strange thing is that the velocity of the multi-level case damped much slower than the single-resolution one. (I was using the one-time damping for single resolution as well, as can be verified by the average damping inner neighbor.) If I use the normal inner relation (particle pairs see each other twice), the result is better, but the oscillation is still larger than the single-one which is updated only once.

image
image
image

Left: single-resolution, using inner_split. Middle: two-mesh levels, using inner_split. Right: two-mesh levels, using inner.

@DongWuTUM The inner-neighbors of single and multi should be exactly the same in this case, I don't have any idea about how this difference come from. I'll test on efficiency first, maybe we can discuss it tomorrow

WeiyiVirtonomy:
## Try to update particles at the same level by parallel 

Try to execute damping by two steps:

1. `parallel(mesh_levels_[l]->particle_for_split(par, function1))`
2. `sequential(mesh_levels_[l]->particle_for_split(par, function2))`

Function 1 only executes damping for particles at the same level, while function 2 only for particles at different levels.

```
    Real Vol_i = this->Vol_[index_i];
    Real capacity_i = this->damping_.Capacity(index_i);

    Neighborhood &inner_neighborhood = this->inner_configuration_[index_i];
    for (size_t n = 0; n != inner_neighborhood.current_size_; ++n) // forward sweep
    {
        size_t index_j = inner_neighborhood.j_[n];
        if (skip(index_i, index_j)) // for function 1, level_[index_i] != level_[index_j]
            continue;
        // codes for damping computation
    }
```

The computational time of the following 5 cases is studied:

1. Execute for each mesh level in sequence (the old way)
2. Execute by the method described above
3. Only execute `parallel(mesh_levels_[l]->particle_for_split(par, function1))` (same level only)
4. Replace function2 with void function (access particle data in each cell, but do nothing to them)
5. Replace `skip` in function2 with `[](size_t, size_t){return true;}` (access neighbor data, but do nothing to them)

## Refinement level = 1
image

For ${dp}_0 = t/8$, ${dp}_1 = t/16$, the time is:

| Type | total time | damping time |
|------|------------|-----------------|
| 1 | 24 | 11 |
|2 | 29 | 13 |
| 3 | 22 | 7 |
| 4 | 24 | 9 |
| 5 | 25 | 11 |

The damping time of type 1 and type 5 is very close. The mean computational cost seems to come from data access instead of computation. If we cannot find a better way to restructure the two-step damping update, the computational efficiency will not be enhanced.

WeiyiVirtonomy:
> ## Further check on the cause of high computational time
> I compared the results with a single resolution d p = h / 16 and a multi-resolution d p 0 = h / 8 and d p 1 = h / 16 . The whole domain is the refinement region so that all the particles in the multi-resolution case are of d p 1 = h / 16 . Hence, the particle number and inner configurations of the mr case is nearly the same as the single resolution one.
> 
> image image
> Left: initial particle distribution of the single resolution case Right: initial particle distribution of the multi-resolution case
> 
> For the multi-resolution case, the cell level is set as:
> 
> 1. Only level 0 with (17, 8, 8) cells
> 2. Level 0 with (17, 8, 8) and level 1 with (29 11 11).
> 
> For the second type, all the particles belong to level 1.
> 
> The cell number of the single-resolution case is (29 11 11).
> 
> ## Computation time
> The particle number, inner neighbor number and computational time of the three cases are:
> 
> Type	Particle number	Average inner neighbor	Average damping inner neighbor	Computational time	Damping time
> Single	14336	44	22	37	7
> MR, one level	14326	43	21	63	29
> MR, two level	14326	43	21	39	8
> The damping time of the two level one is very close to the single-resolution one, so we might conclude that one cause of poor efficiency is the poor cell-parallelization when we only use one cell level.
> 
> ## Weird damping behavior
> The strange thing is that the velocity of the multi-level case damped much slower than the single-resolution one. (I was using the one-time damping for single resolution as well, as can be verified by the average damping inner neighbor.) If I use the normal inner relation (particle pairs see each other twice), the result is better, but the oscillation is still larger than the single-one which is updated only once.
> 
> image image image
> Left: single-resolution, using inner_split. Middle: two-mesh levels, using inner_split. Right: two-mesh levels, using inner.
> 
> @DongWuTUM The inner-neighbors of single and multi should be exactly the same in this case, I don't have any idea about how this difference come from. I'll test on efficiency first, maybe we can discuss it tomorrow

The difference in the speed to reach steady-state might come from `AdaptiveInnerRelation`. 
I directly assigned the position by a lattice style to avoid the effect of initial particle distribution. As the initial position of the single-resolution and fully-refined multi-resolution cases are exactly the same, the inner neighbor values should also be the same, whereas the highest $abs({dW}_{ij}/r_{ij})$ in the inner neighbor of the previous one is 1.9e12, while the later one is only 4.8e11.

I'll go on to check if the problem comes from adaptive search length or neighbor builder.

Xiangyu-Hu:
> I got it, it is this function `Kernel::dW(const Real &h_ratio, const Real &r_ij, const Vec3d &displacement)`, it should be `h_factor_dW_3D_(h_ratio)` but mistakenly typed as `h_factor_dW_1D_(h_ratio)`
> 
> After correction: image

Great! So the puzzle solved?

WeiyiVirtonomy:
> > I got it, it is this function `Kernel::dW(const Real &h_ratio, const Real &r_ij, const Vec3d &displacement)`, it should be `h_factor_dW_3D_(h_ratio)` but mistakenly typed as `h_factor_dW_1D_(h_ratio)`
> > After correction: image
> 
> Great! So the puzzle solved?

Yes, now the result is consistent with the single-resolution ones.

There is still some problem with the body shape. When using the lattice generator, the multi-resolution one had some voids at the corners:
image

After relaxation, the edges are also less smooth as the single-resolution one:
image

The initial shape is the same, so there might be something wrong with the level-set shape definition.

Xiangyu-Hu:
@WeiyiVirtonomy  Yes. Please have a check on level set profile.

Xiangyu-Hu:
> @WeiyiVirtonomy Yes. Please have a check on level set profile.

May be also related to this issue.

WeiyiVirtonomy:
> @WeiyiVirtonomy Yes. Please have a check on level set profile.

The level set is fine, it's just I didn't use it correctly.

`defineBodyLevelSetShape()` should be executed after `defineAdaptation<ParticleRefinementWithinShape>()`, otherwise the grid spacing is not defined according to the minimal spacing. If it is called before the adaptation definition, the input variable should be refinement_level + 1: `defineBodyLevelSetShape(refinement_level + 1)`.

After fixing:
image

DongWuTUM:
> > @WeiyiVirtonomy Yes. Please have a check on level set profile.
> 
> The level set is fine, it's just I didn't use it correctly.
> 
> `defineBodyLevelSetShape()` should be executed after `defineAdaptation<ParticleRefinementWithinShape>()`, otherwise the grid spacing is not defined according to the minimal spacing. If it is called before the adaptation definition, the input variable should be refinement_level + 1: `defineBodyLevelSetShape(refinement_level + 1)`.
> 
> After fixing: image

Okay, great :)

Xiangyu-Hu (approved):

WeiyiVirtonomy:
@Xiangyu-Hu Would it be possible to revert the merge? I need to clean up the code before merging

PR #642, opened 2024-08-23
Rkcg for open boundary flow pull request

FengWang3119:
1. Make the RKCG applicable for OpenBoundaryFlow by Correcting Bj when j is truncated.
2. Correct the bug that in the DensitySummationFreestream, the volume is not updated.
3. Improve the pressure outlet to make it compatible with RKCG.

### DIFF
--- a/src/shared/particle_dynamics/fluid_dynamics/fluid_integration.hpp
+++ b/src/shared/particle_dynamics/fluid_dynamics/fluid_integration.hpp
@@ -84,7 +84,7 @@ void Integration1stHalf<Inner<>, RiemannSolverType, KernelCorrectionType>::inter
         Real dW_ijV_j = inner_neighborhood.dW_ij_[n] * Vol_[index_j];
         const Vecd &e_ij = inner_neighborhood.e_ij_[n];
 
-        force -= (p_[index_i] * correction_(index_j) + p_[index_j] * correction_(index_i)) * dW_ijV_j * e_ij;
+        force -= (p_[index_i] * correction_(index_j, index_i) + p_[index_j] * correction_(index_i)) * dW_ijV_j * e_ij;
         rho_dissipation += riemann_solver_.DissipativeUJump(p_[index_i] - p_[index_j]) * dW_ijV_j;
     }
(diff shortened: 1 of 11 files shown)


src/shared/particle_dynamics/fluid_dynamics/fluid_integration.hpp
Bo-Zhang1995, line 1: the line 87 why the correction for pressure_i using index_i and index_j?

Bo-Zhang1995: correction(index_j, index_i) will return B_[index_j], what is the difference against correction(index_j)?

src/shared/particle_dynamics/particle_functors.h
```diff
@@ -199,13 +204,41 @@ class LinearGradientCorrection : public KernelCorrection
...
+class LinearGradientCorrectionWithBulkScope : public KernelCorrection
+{
+  public:
+    LinearGradientCorrectionWithBulkScope(BaseParticles *particles)
+        : KernelCorrection(),
+          B_(*particles->getVariableDataByName<Matd>("LinearGradientCorrectionMatrix")),
+          within_scope_(particles){};
+
+    Matd operator()(size_t index_j, size_t index_i)
+    {
```
Bo-Zhang1995, line 230: I got it. This is the answer to my concern. There is a selection between i and j.

Xiangyu-Hu: This is not satisfactory. However, at present no other good way to incorporate the change.
Actually, this case raise a issue on how to handle boundary like region within a body.  it is not other body, not ghost and need to handle separately,

src/shared/particle_dynamics/fluid_dynamics/fluid_integration.hpp
FengWang3119: For ordinary case where j is not truncated, the result is the same.

Xiangyu-Hu:
@FengWang3119 why you close the pull request?

PR #88, opened 2022-02-17
Merge upstream/pr 2022.02.16

FabienPean-Virtonomy:
This PR attempts to reconcile the two repositories. We will work on the sync process on our side to avoid a difficult merge of this scale in the future. This should lead to less headache and file conflicts for everyone during merging.

Files with specific need for review
* [in_output.h](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-4f3cb9b96ad37e00398725e68dd75f8b924e240b3b1b21526261826ac39b534c)
* [in_output.cpp](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-116def3b0e2a41c0a70d501aa889702e096bbb157d3fc336cc4d154efc105879)
* [elastic_solid.cpp](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-531ffe9ea5d26e3436b68f074d23b748638adcb19d31c965fa693321c9f82576)
* [elastic_solid.h](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-8a548bb71ba67207ed10c97e4931ad884fec3a881fdb43f6a1075532a9ee934f)
* [contact_dynamics.cpp](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-a3de9b3ddabed276b4327f9e150397f1c3b82a4173abc173d041529157b7d724)
* [contact_dynamics.h](https://github.com/Xiangyu-Hu/SPHinXsys/compare/master...BenceVirtonomy:merge_upstream/PR_2022.02.16#diff-4de3a1f3435f4f213e3f8791e7d57364688668441180a8aac7b729a3f52f4b8f)

Tests removed from ctest due to issues with CI
* test_2d_self_contact
* test_2d_sliding
* test_2d_dambreak
* test_3d_taylor_bar

@BenceVirtonomy, if you have anything to add here

PR #643, opened 2024-08-23
T-shaped pipe of VIPO with shell

ChenxiZhaoTUM:
This is a test of fluid shell interaction in a T-shaped pipe.
The flow with velocity inlet and pressure outlet boundary condition in a T-shaped pipe in 2D is considered.


ChenxiZhaoTUM:
> @ChenxiZhaoTUM could you post some results here so we can have a look? thanks.

Here are the results of velocity and pressure force distributions.

Output step 10:
t10

Output step 15:
t15

Output step 150:
t150

Xiangyu-Hu:
@ChenxiZhaoTUM the case seems quite small. Why does take so long time to finish?
88/160 Test  #88: test_2d_T_pipe_VIPO_shell ......................................   Passed  488.79 sec

Xiangyu-Hu (approved):

PR #647, opened 2024-09-02
try to implement mesh dynamics

Nafug22:
wrap functions into kernels (mesh dynamics) for later use.


Xiangyu-Hu:
@Franguuuu Could you also add brief description of the changes.

PR #658, opened 2024-09-15
Xiangyu/revise bidirectional buffer

Xiangyu-Hu:
1. introduced body part index so that one can identify different buffer regions.
2. do particle deletion within the bi-directional buffer class with checking on body part index.
3. add comment that we need first inject for all buffers then do deletion for them.
4. found the mixing between injection and deletion in parallel is dangerous.


src/shared/bodies/base_body.cpp
```diff
@@ -36,6 +36,12 @@ BoundingBox SPHBody::getSPHSystemBounds()
     return sph_system_.system_domain_bounds_;
 }
 //=================================================================================================//
+int SPHBody::getNewBodyPartID()
```
FengWang3119, line 39: Checking this ID seems to be better and easier than using the checkNotFar.  
But I guess there is a potential bug for this, if the .tag_buffer_particles.exec(); is in front of the deletion.
Since the outflow particles would not be tagged as the buffer particle.

tests/extra_source_and_tests/test_2d_T_pipe_VIPO_shell/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ ADD_EXECUTABLE(${PROJECT_NAME} ${EXECUTABLE_OUTPUT_PATH} ${DIR_SRCS})
 add_test(NAME ${PROJECT_NAME} COMMAND ${PROJECT_NAME} --state_recording=${TEST_STATE_RECORDING}
     WORKING_DIRECTORY ${EXECUTABLE_OUTPUT_PATH})
 
-set_tests_properties(${PROJECT_NAME} PROPERTIES LABELS "fluid_shell_interaction_VIPO")
+set_tests_properties(${PROJECT_NAME} PROPERTIES LABELS "pressure boundary")
```
FengWang3119, line 23: Just curious what is the function of changing the CMakeList file?

src/shared/bodies/base_body.cpp
Xiangyu-Hu: Yes. Tagging need to done at initial condition and immediately after cell linked list is done.

tests/extra_source_and_tests/test_2d_T_pipe_VIPO_shell/CMakeLists.txt
Xiangyu-Hu: This a ctest parameter. With argument -L and the key word such as pressure,  ctest wil only test all the tests with this key word.

FengWang3119 (approved):

PR #659, opened 2024-09-16
Xiangyu/combining injection and deletion

Xiangyu-Hu:
Combine particle injection and deletion  in bi-drectional buffer so that it  is identified from particle disposer class.


Xiangyu-Hu:
I decide to close the pull request, as it seems not a good idea to mix the copy direction.
That is mixing the copy to and from the end of real particles, which is quite dangerous for memory conflict issue.
May be the better way to improve efficiency is to combine the injection or deletion in all buffers together in corresponding single loop.

PR #673, opened 2024-10-06
Xiangyu/buffer consistency

Xiangyu-Hu:
The particle injecting only after the entire inflow particle leave the buffer. So, the consistency can be achieved.


FengWang3119:
QQ_1728205390196
This is the paper I mentioned, I think they also find this problem.
http://dx.doi.org/10.1016/j.cpc.2016.09.009

Xiangyu-Hu:
@FengWang3119 Could you approve the pull request if there is no more issue? Thanks.

FengWang3119 (approved):

PR #675, opened 2024-10-12
Xiangyu/sycl loop ranges

Xiangyu-Hu:
Using loop range class to generalize the scope of compute kernels-


Xiangyu-Hu:
I am not this is necessary.

Issue #469, opened 2023-11-11
Crash with without error message when regression data not written.


Xiangyu-Hu:
The regression test class is defined but not used to output data.
For example in the water exit case, one define the test:
https://github.com/Xiangyu-Hu/SPHinXsys/blob/d85a5e6683947acd1d7964a636a6bcb1bd6b79d4/tests/2d_examples/test_2d_water_entry_exit/test_2d_water_entry_exit.cpp#L392-L393
But forget to write the output file, and still try to generate the regression test data by:
https://github.com/Xiangyu-Hu/SPHinXsys/blob/d85a5e6683947acd1d7964a636a6bcb1bd6b79d4/tests/2d_examples/test_2d_water_entry_exit/test_2d_water_entry_exit.cpp#L538
When I run the python script, I obtained crash information, but not able find the reason, 
that I actually have not write out data.
Therefore, we need an error handler so that one can notice that the data file was not written.

Bo-Zhang1995:
I will have a look on that.

Xiangyu-Hu:
Another issue is that there is repeating data, the the reading the xml lead to error and and the regression data can nor be generated.

Xiangyu-Hu:
@Bo-Zhang1995 Please look the these bugs.

Bo-Zhang1995:
sry forget the previous bug. i will address these asap.- - - - - -张博On 28 Jan 2024, at 13:06, Xiangyu Hu ***@***.***> wrote:﻿
@Bo-Zhang1995 Please look the these bugs.

Xiangyu-Hu:
@Bo-Zhang1995 How about this bug?

Bo-Zhang1995:
This means that the input/output scheme of the database should be optimized, as we discussed during lunch.
The existing database will result in no new data being generated. I think this could be optimized by adding some alerts to warn when the data is repeated in the database. I will take some time to optimize it. Sorry for that the last time I was in China.

Xiangyu-Hu:
done with check when initializing regression test:
https://github.com/Xiangyu-Hu/SPHinXsys/blob/master/src/shared/regression_test/regression_test_base.hpp

PR #682, opened 2024-11-11
regression test problem #469

Bo-Zhang1995:
Here, the database will be emptied for each new generation. Therefore, this problem can be solved.


Xiangyu-Hu:
@Bo-Zhang1995 Could you modify the pull request so that all tests can pass? Thanks.

Xiangyu-Hu (approved):

PR #685, opened 2024-11-20
transport velocity in soil

Shuaihao-Zhang:
Remove the tensile instability in cohesive soil using transport velocity.
image


tests/extra_source_and_tests/test_2d_cohesive_soil_failure/cohesive_soil_failure.h
```diff
@@ -0,0 +1,307 @@
...
+        Real stress_yy = -rho0_s * gravity_g * y;
+        stress_tensor_3D_[index_i](1, 1) = stress_yy;
+        stress_tensor_3D_[index_i](0, 0) = stress_yy * gama;
+        stress_tensor_3D_[index_i](2, 2) = stress_yy * gama;
+    };
+};
+//----------------------------------------------------------------------
+//	Unified transport velocity correction
+//----------------------------------------------------------------------
+template <typename... T>
```
Xiangyu-Hu, line 86: I am wondering why we need this new class. Could we use a generalization-extension approach to add this new feature?

Shuaihao-Zhang: > I am wondering why we need this new class. Could we use a generalization-extension approach to add this new feature?

Yes. It's better to use a generalized approach.
A simple approach is to just add a judgment statement in the original `TransportVelocityCorrection`.
When a particle is identified as a surface particle, corrections are applied only in the tangential direction. However, it is uncertain whether this approach will affect other cases.

Xiangyu-Hu: No. Using a switch is not a generalization-extension approach. We need first find a type in the original formulation such that it can be generalized, then change it to a template parameter (as a templated class) and reimplement the original class by realizing the template parameter with a small function object. For extension, we write the new feature as a new function object which substitute the template parameter in the generalized templated class. In this way, the choose new feature is don in compiling time not runtime.

Shuaihao-Zhang: > No. Using a switch is not a generalization-extension approach. We need first find a type in the original formulation such that it can be generalized, then change it to a template parameter (as a templated class) and reimplement the original class by realizing the template parameter with a small function object. For extension, we write the new feature as a new function object which substitute the template parameter in the generalized templated class. In this way, the choose new feature is don in compiling time not runtime.

I will think about this.

Xiangyu-Hu:
@Shuaihao-Zhang the bug in sycl is solved already, you need pull the master first and update he pull request.

Shuaihao-Zhang:
> @Shuaihao-Zhang the bug in sycl is solved already, you need pull the master first and update he pull request.

I have merged the master into the current branch.

Shuaihao-Zhang:
@Xiangyu-Hu All tests have passed.

Xiangyu-Hu (approved):

PR #735, opened 2025-01-30
Ck sycl/kernel summation ck

YuVirtonomy:
This pull request introduces new functionalities and enhancements to the fluid dynamics module of the SPHinXsys library. The main changes include the addition of new header files for kernel summation and transport velocity correction, along with their corresponding implementations. These updates aim to improve the simulation accuracy and extend the capabilities of the fluid dynamics computations.

### New Functionalities:

* Added `kernel_summation_ck.h` and its implementation to define algorithms for computing density using kernel function summation. This includes template classes for different interaction types and their specializations. [[1]](diffhunk://#diff-9a3bbc3c0da901080784e5e46cbcb5ce06af0c9b0c8fbc9fc27c8796eb9d0097R1-R147) [[2]](diffhunk://#diff-021e13ba5118a90629a449c2f7e101f053a3472df1b0125355069dae214cf5f9R1-R105)
* Added `transport_velocity_correction_ck.h` and its implementation to introduce transport velocity correction for different particle scopes. This includes template classes and their specializations for various particle types and interaction scenarios.

### Header File Inclusions:

* Included new header files `kernel_summation_ck.hpp` and `transport_velocity_correction_ck.hpp` in `all_shared_fluid_dynamics_ck.h` to ensure they are part of the shared fluid dynamics module.

### Result
The kernel summation method was implemented and validated using a pipe flow scenario in the original implementation. However, we currently lack a dedicated emitter class for CK. As a temporary workaround, we used a lid-driven cavity case for testing.

Commit [625420a4763d9d56d1c1bb58795556466a62ef5f](https://github.com/user/repo/commit/625420a4763d9d56d1c1bb58795556466a62ef5f) presents a comparison of results with and without wall contact when applying kernel summation.

- The left side of the image represents the case without wall contact. As expected, the kernel summation values are higher near the wall.
- The right side of the image represents the case with wall contact, where the kernel summation values near the wall are closer to zero.
 
This behavior aligns with theoretical expectations, as kernel summation approaches zero when no particles are missing within its support domain. This demonstrates that the method accurately accounts for the presence of walls in the simulation.
image
image
image


Xiangyu-Hu:
@YuVirtonomy could you explain a bit on the changes?

YuVirtonomy:
> @YuVirtonomy could you explain a bit on the changes?

Refactored Kernel Summation Handling in TransportVelocityCorrectionCK
1.     Unified kernel summation (NablaWV_) with zero_gradient_residue_, reducing redundant loops.
2.     Added NablaWVSummationType as a template parameter for compile-time optimization.
3.     Used if constexpr to eliminate unnecessary calculations when WithoutKernelSummationCK is used.

I think it might be good to keep the original kernel summation ck files and remove it later on if its really become redundant

Xiangyu-Hu:
I am wondering why we need one extra template parameter. Could we just use variable name indicating meaning?

YuVirtonomy:
After our discussion, we have decided to drop kernel_summation and replace it with zero_gradient_residue_[index_i]. An important aspect of this change is that we will need to adjust how the momentum equation is applied.

Current Implementation

In the original pressure_boundary.h, the momentum equation update is applied as follows:
`vel_[index_i] += 2.0 * kernel_sum_[index_i] * target_pressure_(p_[index_i], *physical_time_) / rho_[index_i] * dt;`

Proposed Change

With the replacement of kernel_summation by zero_gradient_residue_, the updated equation will be:
`vel_[index_i] += zero_gradient_residue_[index_i] * target_pressure_(p_[index_i], *physical_time_) / rho_[index_i] * dt;`

Note: The factor 2.0 is already incorporated within zero_gradient_residue_, so it is no longer needed in the equation

As this change is bigger scope than this PR, we will closed this PR once this mentioned feature is done with pressureboundary CK version.

YuVirtonomy:
As discussed earlier, this branch is redundant since zero_gradient_residues can fully replace kernel summation. Additionally, the feature has already been implemented in [PR #756](https://github.com/Xiangyu-Hu/SPHinXsys/pull/756).

Therefore, this PR is no longer needed and will be closed.

Issue #743, opened 2025-02-09
Seperate template classes for 2d and 3d anisotropic diffusion cases.


Xiaojingtang1234:
For anisotropic diffusion algorithm,  in 2d case,` vec2d , Mat3d , Vec3d ` variables exist at the same time, which means the corresponding `AnisotropicKernelCorrectionMatrix` and `AnisotropicDiffusionRelaxation`  can only  be put in **for_2d_build** and  **for_3d_build** folder instead of  the **shared folder**. Otherwise, if we use  **vecd** in these template classes,  the compiler will recognize all **vecd**  as **vec2d**  by default in 2d case, where **vec3d**  exist as well.
Therefore, we need seperate template classes for 2d and 3d cases.@Xiangyu-Hu

Xiangyu-Hu:
No. The present code allow 2d and 3d variable in any class at the same time. Also, the AnisotropicKernel can be implemented in a single file using vector and matrix operations.

Xiaojingtang1234:
The question is:
In 3d case: **mat6d** and **mat3d** are both used.
In 2d case: **mat3d** and **mat2d** are both used.
We also need explicit definetions of some variables like:  
`Vec3d disp_quad_ = Vec3d(r_ji[0] * r_ji[0], r_ji[1] * r_ji[1], r_ji[0] * r_ji[1]);` in 2d case.
`Vec6d disp_quad_ = Vec6d(r_ij[0] * r_ij[0], r_ij[1] * r_ij[1], r_ij[2] * r_ij[2], r_ij[0] * r_ij[1], r_ij[1] * r_ij[2], r_ij[2] * r_ij[0]);` in 3d case.
           
Then how should we define variable types in a generalized template class? The compiler might encounter issues when instantiating the variable type.

Xiangyu-Hu:
[...]

Issue #742, opened 2025-02-09
add displacement rij


Xiaojingtang1234:
For anisotropic diffusion algorithm, two displacements are needed, one is the real displacment, and the other one is the transformed displacement.  Therefore, I suggest  that we add another variable` StdLargeVec<Vecd> r_ij_vector_;` in `class Neighborhood`.
@Xiangyu-Hu 

```void NeighborBuilder::createNeighbor(Neighborhood &neighborhood, const Real &distance,
const Vecd &displacement, size_t index_j)
{
    neighborhood.r_ij_.push_back(distance);
    neighborhood.r_ij_vector_.push_back(displacement);
}
//=================================================================================================//
void NeighborBuilder::initializeNeighbor(Neighborhood &neighborhood, const Real &distance,
                                         const Vecd &displacement, size_t index_j)
{
    neighborhood.r_ij_[current_size] = distance;
    neighborhood.r_ij_vector_[current_size] = displacement;
}
```

Xiangyu-Hu:
No. It think that the present neighborhood is too memory intensive already. If the value can be computed on the fly just compute it when use it.

PR #744, opened 2025-02-09
Anisotropic diffusion

Xiaojingtang1234:
This is for the anisotropic diffusion coupling with anisotropic resolution.

- [x] variables `vec6d, mat6d` are implemented in **base_data_type.h**
- [x] **anisotropic_diffusion_relaxation.hpp** is created for template classes.
- [x] `AnisotropicKernelCorrectionMatrix` is for the further kernel correction.
- [x] ` AnisotropicDiffusionRelaxation ` is for the anisotropic diffusion relaxation.`*Laplacian_x, *Laplacian_y,` are recorded for checking the diffusion rate in different directions.
- [x] The results are validated with functions    $$\Phi = x +y $$  and  $$\Phi = x^2 +y^2 $$.
 

Note that two issues related are  created. 
1, The displacement issue should be implemented another real **rij_vector.** . Related to #742
2, The current 2d case is depend on  **anisotropic_diffusion_relaxation.hpp** . Related to #743 

Additionaly,  the neighborhood statement should be modified to **if (kernel_->checkIfWithinCutOffRadius(displacement) )**, matching the one in **NeighborBuilderInner**, note that `index_i != index_j ` should be deleted, considering index maybe the same in different bodies.
```
void NeighborBuilderContact::operator()(Neighborhood &neighborhood,
                                        const Vecd &pos_i, size_t index_i, const ListData &list_data_j)
{
    size_t index_j = list_data_j.first;
    Vecd displacement = pos_i - list_data_j.second;
    Real distance = displacement.norm();
   if (kernel_->checkIfWithinCutOffRadius(displacement) )
    {
        neighborhood.current_size_ >= neighborhood.allocated_size_
            ? createNeighbor(neighborhood, distance, displacement, index_j)
            : initializeNeighbor(neighborhood, distance, displacement, index_j);
        neighborhood.current_size_++;
    }
};
```


Xiaojingtang1234:
Issue 742 can be closed then with  rij  being calculated in interaction part, no need to do any change in base code. Solved https://github.com/Xiangyu-Hu/SPHinXsys/issues/742

Xiaojingtang1234:
The CI test of **test_2d_anisotropic_beam** cannot pass because if we fix the bug in **NeighborBuilderContact::operator()** , (see the last comment) , the regression test data of this case involving anisotropic kernel needs to be updated. @Xiangyu-Hu

Xiangyu-Hu:
> The CI test of **test_2d_anisotropic_beam** cannot pass because if we fix the bug in **NeighborBuilderContact::operator()** , (see last comment) , the regression test data of this case involving anisotropic kernel needs to be updated. @Xiangyu-Hu 

Why needs to be updated?

Xiaojingtang1234:
> > The CI test of **test_2d_anisotropic_beam** cannot pass because if we fix the bug in **NeighborBuilderContact::operator()** , (see last comment) , the regression test data of this case involving anisotropic kernel needs to be updated. @Xiangyu-Hu
> 
> Why needs to be updated?

Firstly there is a bug that: 
The previous statement in **NeighborBuilderContact::operator()**  : if (distance < kernel_->CutOffRadius())
Updated one:   if (kernel_->checkIfWithinCutOffRadius(displacement) ).

The contact neighbourhood is determined by this statement, which does not impact the simulation result since observation in not involved into the simuation. 
But the regression test data is generated from the observe body, where the kernel is reset to **AnisotropicKernel<KernelWendlandC2>**. 
If we fix the bug, the  previous measurement of displacement is not accurate anymore. 
That is why we need to update the data base.
```
 RegressionTestEnsembleAverage<ObservedQuantityRecording<Vecd>>
        write_beam_tip_displacement("Position", beam_observer_contact);
```
   ```
 ObserverBody beam_observer(system, "BeamObserver");
    beam_observer.sph_adaptation_->resetKernel<AnisotropicKernel<KernelWendlandC2>>(scaling_vector);
```

**Comment by Xiaojingtang1234** [comment]
All checks have passed. A review is needed. @Xiangyu-Hu

File: tests/extra_source_and_tests/extra_src/for_2D_build/anisotropic_diffusion_relaxation.hpp [hunk 1]
```diff
@@ -0,0 +1,446 @@
...
+
+  protected:
+    Real *Vol_;
+    Matd *B_;
+    Vecd *pos_;
+
+};
+
+template <>
+class AnisotropicLinearGradientCorrectionMatrix<Inner<>>
```
Xiangyu-Hu, line 62: This exact the same as B computed in the library. I do not think that we need a new class for it.

tests/extra_source_and_tests/extra_src/for_2D_build/anisotropic_diffusion_relaxation.hpp
```diff
@@ -0,0 +1,446 @@
...
+        Matd inverse = (B_T * B_[index_i] + SqrtEps * Matd::Identity()).inverse() * B_T;
+        B_[index_i] = inverse ;
+    };
+    
+
+};
+using AnisotropicLinearGradientCorrectionMatrixInner = AnisotropicLinearGradientCorrectionMatrix<Inner<>>;
+
+template <>
+class AnisotropicLinearGradientCorrectionMatrix<Contact<>>
```
Xiangyu-Hu, line 101: Again, it is exact the same the one in the main library.

tests/extra_source_and_tests/extra_src/for_2D_build/anisotropic_diffusion_relaxation.hpp
```diff
@@ -0,0 +1,446 @@
...
+   
+    Vec2d *kernel_correction1_,*kernel_correction2_, *kernel_correction3_;
+    Real *Vol_;
+    Mat2d *B_;
+    Vecd *pos_;
+
+}; 
+
+template <>
+class AnisotropicKernelCorrectionMatrix<Inner<>>
```
Xiangyu-Hu, line 173: This is only for 2d case.

Xiangyu-Hu:
@Xiaojingtang1234 I think that we need write the formulation together for 2 and 3d, and we need a 3D case.

tests/extra_source_and_tests/extra_src/for_2D_build/anisotropic_diffusion_relaxation.hpp
Xiangyu-Hu, line 1: We need write the correction in the same file for 2 and 3d correction and we need 3d case.

tests/extra_source_and_tests/extra_src/for_2D_build/anisotropic_diffusion_relaxation.hpp
Xiaojingtang1234: That is what we talked before, a real anisotropic **rij** is needed here. But in the current version, the rij is obtained from neighbourhood, which has been tranfromed to the isotropic one.
Main library: ` Vecd r_ji = contact_neighborhood.r_ij_[n] * contact_neighborhood.e_ij_[n];`
New class:    `Vecd r_ji = pos_[index_j] - pos_[index_i];`
They are different in the current library. If we want make them the same, the expresssions in kernel need to be modified.

Xiaojingtang1234:
Three dimensional case has been set. I tried, but the code still looks ugly.

PR #765, opened 2025-03-04
Corrected interpolation with restoring matrix

Bo-Zhang1995:
For the interpolation correction, a large restoring matrix needs to be constructed for both 2D and 3D cases. 
Using the Eigen block can implement this function.

Currently, just another interpolation is added, and it can be replaced by the template in the future.


Bo-Zhang1995:
The finite particle method is used for correcting the interpolation. 2D and 3D are unified version.

Xiangyu-Hu:
@Bo-Zhang1995 is this work still in progress?

Bo-Zhang1995:
@Xiangyu-Hu Here, I redefine the interpolation as BaseInterpolationCorrected. Do you think it would be better to separately redefine InterpolatingAQuantity and ObservingAQuantity, or should I introduce an argument to distinguish between corrected and non-corrected interpolation?
If I use an additional template parameter for these two classes, ObservedQuantityRecording would also need to be refined with an extra template argument. Do you have any suggestions for a cleaner or more efficient approach?

Xiangyu-Hu:
> @Xiangyu-Hu Here, I redefine the interpolation as BaseInterpolationCorrected. Do you think it would be better to separately redefine InterpolatingAQuantity and ObservingAQuantity, or should I introduce an argument to distinguish between corrected and non-corrected interpolation? If I use an additional template parameter for these two classes, ObservedQuantityRecording would also need to be refined with an extra template argument. Do you have any suggestions for a cleaner or more efficient approach?

I am not sure. For now, I think that you just need do as the original code is doing.

Bo-Zhang1995:
> > @Xiangyu-Hu Here, I redefine the interpolation as BaseInterpolationCorrected. Do you think it would be better to separately redefine InterpolatingAQuantity and ObservingAQuantity, or should I introduce an argument to distinguish between corrected and non-corrected interpolation? If I use an additional template parameter for these two classes, ObservedQuantityRecording would also need to be refined with an extra template argument. Do you have any suggestions for a cleaner or more efficient approach?
> 
> I am not sure. For now, I think that you just need do as the original code is doing.

I just added another interpolation and observation. They can be rearranged in the future.

src/shared/particle_dynamics/general_dynamics/general_interpolation.h
```diff
@@ -78,11 +78,88 @@ class BaseInterpolation : public LocalDynamics, public DataDelegateContact
...
+                Real dW_ij = contact_neighborhood.dW_ij_[n];
+
+                Real element1 = W_ij * Vol_k[index_j];
+                Vecd element2 = W_ij * Vol_k[index_j] * r_ji;
+                Vecd element3 = dW_ij * Vol_k[index_j] * e_ij;
+                Matd element4 = dW_ij * Vol_k[index_j] * r_ji * e_ij.transpose();
+
+                prediction(0, 0) += element1 * data_k[index_j];
+                for (Eigen::Index i = 1; i < prediction.rows(); ++i) {
+                    prediction(i, 0) += element3[i] * data_k[index_j];
```
Xiangyu-Hu, line 142: something wrong

PR #766, opened 2025-03-05
add an option which allowed you to build sphinxsys_2d or sphinxsys_3d with files in extra_src folder

yyc84:
Add an option in CMakelist that allows you to build sphinxsys_2d or sphinxsys_3d with files in the extra_src folder.
By clicking "SPHINXSYS_USE_EXTRA_SOURCES" button in the CMake GUI  
One can use this function to test some temporary code with original source files (imaging some classes need to include the new header in your extra_src folder to test the new function)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,6 +6,9 @@ file(GLOB_RECURSE SPHINXSYS_SYCL_HEADERS CONFIGURE_DEPENDS src_sycl/*.h src_sycl
 file(GLOB_RECURSE SPHINXSYS_SYCL_SOURCES CONFIGURE_DEPENDS src_sycl/*.cpp)
 endif()
 
+# **add an option to choose whether include th extra soure files or not**
+option(SPHINXSYS_USE_EXTRA_SOURCES "Include extra source files in sphinxsys_2d and sphinxsys_3d" OFF)
+
 if(SPHINXSYS_2D)
     add_library(sphinxsys_2d)
     #Make sure in-tree projects can reference this as SPHinXsys::sphinxsys_2d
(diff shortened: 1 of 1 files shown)


Xiangyu-Hu:
@yyc84 Do you only 2d or 3d code in extra... folder?

Xiangyu-Hu:
@yyc84 I think that currently the cases in extra... will be build and tested.
as
https://github.com/Xiangyu-Hu/SPHinXsys/blob/master/tests/extra_source_and_tests/extra_src/CMakeLists.txt

yyc84:
> @yyc84 Do you only 2d or 3d code in extra... folder?

and this is only an option in CMakelist; if you choose then you can use the original library (sphinxsys_2d or 3d) to test your new function in the extra_src folder. it won't affect the original code

Xiangyu-Hu:
> > @yyc84 I think that currently the cases in extra... will be build and tested. as https://github.com/Xiangyu-Hu/SPHinXsys/blob/master/tests/extra_source_and_tests/extra_src/CMakeLists.txt
> > Now, it will be build and test by default. You change it as an option?
> 
> no, the reason I modified this cmakelikst file is I tested my own new functions, which I put into the extra_src folder, but some functions need to be based on the original classes in the src folder, but the files in the src folder can not find the files in extra_src folder. For example, I am testing the level_set boundary, I creat most of those functions and classes in new files and put into extra_src folder, but I need a function in the "adaptation.h" file to creat a different LevelSet with new functions, so I don't want to creat new adaptation class, only need its function, that is why I need "adaptation.h" to found new files in extra_src folder

I am confused. The present setup allows files in extra... any any functions in src folder. Am I wrong somewhere?

Xiangyu-Hu:
> > > > @yyc84 I think that currently the cases in extra... will be build and tested. as https://github.com/Xiangyu-Hu/SPHinXsys/blob/master/tests/extra_source_and_tests/extra_src/CMakeLists.txt
> > > > Now, it will be build and test by default. You change it as an option?
> > > 
> > > 
> > > no, the reason I modified this cmakelikst file is I tested my own new functions, which I put into the extra_src folder, but some functions need to be based on the original classes in the src folder, but the files in the src folder can not find the files in extra_src folder. For example, I am testing the level_set boundary, I creat most of those functions and classes in new files and put into extra_src folder, but I need a function in the "adaptation.h" file to creat a different LevelSet with new functions, so I don't want to creat new adaptation class, only need its function, that is why I need "adaptation.h" to found new files in extra_src folder
> > 
> > 
> > I am confused. The present setup allows files in extra... use any functions in src folder. Am I wrong somewhere?
> 
> yes, you are right, the present setup allows files in extra_src to use any functions in src folder, but not allow any files in src to use the function in extra_src folder, the header in extra_src can not be found by files in src folder.

In this case, we need more these function case to main src folder_

Xiangyu-Hu:
this property seems not necessary.

PR #767, opened 2025-03-06
add 6d data type

Xiaojingtang1234:
Vec6d and Mat6d should be added into the datatype.


Xiaojingtang1234:
6d data type should be added. @Xiangyu-Hu

Xiangyu-Hu:
@Xiaojingtang1234 we need a case use the new type.

Xiangyu-Hu:
due to pull request #798.

PR #774, opened 2025-03-17
Implicit relaxation method

Bo-Zhang1995:
The implicit relaxation based on the projection is integrated. The current method doesn't involve the contact for complexrelation due to the split dynamics.


tests/2d_examples/test_2d_fsi2/fsi2.cpp
Xiangyu-Hu, line 1: @Bo-Zhang1995 Please open a new test case, other than modify old ones. We will consider to rearrange the cases later.

tests/2d_examples/test_2d_implicit_relaxation/implicit_relaxation.cpp
```diff
@@ -0,0 +1,124 @@
+/**
+ * @file fsi2.cpp
```
Xiangyu-Hu, line 2: These comments are not relevant to case. Also, I prefer a case (4.1.3) in the paper with observation and regression test on kinetic energy.

Xiangyu-Hu (approved):

PR #775, opened 2025-03-17
spawn particles parallel policy

Xiangyu-Hu:
Now, parallel policy for spawn particles

PR #776, opened 2025-03-18
Despawn particles parallel policy

YuVirtonomy:
This PR builds upon #775 (Following spawn particles parallel policy) by implementing a parallel execution policy for despawning particles.

Previously, BidirectionalConditionCK required both ParallelExecutionPolicy and SequencedExecutionPolicy—the former applied to non-spawn and non-despawn features, while the latter applied to spawn and despawn features.

Since parallel execution is now available for both spawn and despawn, SequencedExecutionPolicy in BidirectionalConditionCK has been removed.

### DIFF
--- a/src/shared/common/base_data_type.h
+++ b/src/shared/common/base_data_type.h
@@ -43,6 +43,8 @@
 #if SPHINXSYS_USE_SYCL
 #include <CL/sycl.hpp>
 #define SYCL_DEVICE_ONLY
+#else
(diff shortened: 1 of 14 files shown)


Xiangyu-Hu:
@YuVirtonomy deleting particles is much more complicated when it is in parallel.

YuVirtonomy:
Close due to current structure not capable of dealing with parallel despwan.

PR #777, opened 2025-03-18
check if last real particle is deletable before copy data for deleting a real particle

Xiangyu-Hu:


Xiangyu-Hu:
Find bug first.

PR #778, opened 2025-03-19
add a function to reset the file path by update the quantity name, to…

yyc84:
Add a function in BaseQuantityRecording to reset the file path by updating the quantity name, which can prevent different quantity names from using the same file path when there are several quantities using ReducedQuantityRecording to record the values. This update also has no impact to other classes based on BaseQuantityRecording .

For example:
It was like this in flow around cylinder when you want to record total viscous force and total pressure force together:
In the file, it looks like this:
"run_time"   "TotalViscousForceFromFluid[0]"   "TotalViscousForceFromFluid[1]"   
"run_time"   "TotalPressureForceFromFluid[0]"   "TotalPressureForceFromFluid[1]"   
1.06736   0.129035062   -0.001386581   
1.06736   2.447225204   -0.023816057   
2.07325   0.638323563   -0.006861182

### DIFF
--- a/src/shared/io_system/io_observation.h
+++ b/src/shared/io_system/io_observation.h
@@ -38,10 +38,8 @@ class BaseQuantityRecording : public BaseIO
 {
   public:
     BaseQuantityRecording(SPHSystem &sph_system,
-                          const std::string &dynamics_identifier_name,
-                          const std::string &quantity_name);
-
-    void setQuantityName(const std::string &quantity_name)
+                          const std::string &dynamics_identifier_name);
+    void setFullPath(const std::string &quantity_name)
     {
         quantity_name_ = quantity_name;
         filefullpath_output_ = io_environment_.output_folder_ + "/" +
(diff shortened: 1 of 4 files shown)


Xiangyu-Hu:
@yyc84 Where is the bug them? The original file name is not based quantity name?

src/shared/io_system/io_observation.h
```diff
@@ -41,6 +41,13 @@ class BaseQuantityRecording : public BaseIO
                           const std::string &dynamics_identifier_name,
                           const std::string &quantity_name);
 
+    void setQuantityName(const std::string &quantity_name)
```
Xiangyu-Hu, line 44: Function name better called setFullPath(const std::string &quantity_name).

Xiangyu-Hu: and put this function in base and delete the old lines and delete the construct parameter quantity name for the base class.

yyc84: I remove the parameter "&quantity_name" from the constructor of BaseQuantityRecording, 
modified the function name as "setFullPath" in base class,
set quantity_name a default value as "NeedAQuantityName" in base class
set a default file full path as
"io_environment_.output_folder_ + "/" + dynamics_identifier_name_ + "_" + "NeedAQuantityName" + ".dat") in base class;
remove all the parameter "&quantity_name" from the constructor of all derived class.

But I think it is better to keep two different constructors in Base class, one with "&quantity_name" in constructor, one without "&quantity_name" in constructor.

Xiangyu-Hu (approved):

PR #779, opened 2025-03-22
Xiangyu/sync to cpu formulation

Xiangyu-Hu:
to sync with cpu version of the pressure boundary conditions.


Xiangyu-Hu:
close for now.

PR #817, opened 2025-05-21
Add pressure gradient implementation and corresponding unit test

aerocoder23:
This draft PR includes a custom implementation of pressure gradient calculation using a linear pressure profile (p = 2x + 5) and a corresponding unit test.
I’m currently seeing large errors (~35) instead of the expected gradient value (~2). I would appreciate any feedback before our next meeting.


src/shared/particle_dynamics/fluid_dynamics/pressure_gradient.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+            Real p_j = p_[index_j];
+            Vecd nablaW_ijV_j = contact_neighborhood.dW_ij_[n] * Vol_k[index_j] * e_ij;
+
+            pressure_grad -= (p_i - p_j) * nablaW_ijV_j;
+        }
+    }
+
+    // Duvar için correction YOK, doğrudan sonucu ata
+    p_grad_[index_i] = pressure_grad;
```
Xiangyu-Hu, line 45: Here, you have overwrite the quantity computed by the inner neighbors. 
Should be p_grad_[index_i] += pressure_grad;

tests/unit_tests_src/shared/particle_dynamics/fluid_dynamics/test_2d_pressure_gradient/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 4.0)
```
Xiangyu-Hu, line 1: NO. We do not need 4.0 as it may lead to issues.

PR #844, opened 2025-07-03
access r_i and r_j

Xiangyu-Hu:
small change for access configuration position


Xiangyu-Hu:
It seems that these are not necessary.

PR #848, opened 2025-07-07
Tong wu master project

TongWu602:
Dear Prof.Hu
I have add some new script to make it also work on linux and macos
Thank you very much for helping me test it!
Tong Wu


TongWu602:
Dear Prof. Hu
It seems it had pass all test of three system

PR #852, opened 2025-07-09
PyVista Jupyter notebook visualizer

Victorwithak:
I've finalised the visualizer. I have added the two options of .gif (cross section and full) and some .vtp files to test the jupyter notebook on. I hope I placed it in the correct directory.
I will probably update it this week, since I want to add flow charts to it, but that will come next week.


tests/3d_examples/test_3d_PyVista_Visualizer/PyVista_Visualizer.ipynb
```diff
@@ -0,0 +1,399 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "0dd01b72-fe9a-46fb-8ed8-c4edda84847a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# %pip install pyvista vtk"
```
Xiangyu-Hu, line 22: I think that we need really install the libraries here.

PR #859, opened 2025-07-15
[Discussion] Buffer singularities

WeiyiVirtonomy:
See the 3d t-shape as an example. 
Inlet pressure: 10mmHg
Outlet pressure: 9mmHg
The inlet and outlet pressures increase to the target value linearly.
image
image
image
The velocity inside buffer regions is more irregular and has higher values.


WeiyiVirtonomy:
> @WeiyiVirtonomy And, where is the end of the wall at the inlet. The aligned box surface should be aligned with wall end surface fluid domain end surface.

No extension of wall at the boundary:
image

The aligned box is placed like this:
image

Xiangyu-Hu:
Is the high velocity region always at the same place or it changes location and time?

WeiyiVirtonomy:
> Is the high velocity region always at the same place or it changes location and time?

The location changes over time, but always near the inlet or outlets

Xiangyu-Hu:
@WeiyiVirtonomy you may also try the new pull request #870 and have a look if the issue is due to deleting particle.

Xiangyu-Hu:
> Collaborator Author

Did you check the results?

WeiyiVirtonomy:
> > Collaborator Author
> 
> Did you check the results?

Yes, with the lifecycle changes, the simulation result doesn't show much difference from the previous ones

Xiangyu-Hu:
> > > Collaborator Author
> > 
> > 
> > Did you check the results?
> 
> Yes, with the lifecycle changes, the simulation result doesn't show much difference from the previous ones

Ok. Let's check what the original CPU version can get.

Xiangyu-Hu:
@WeiyiVirtonomy yes. I think that the problem is really originated from linear correction. However, i member that we have already introduced a solution before.  Let us discuss it tomorrow.

Xiangyu-Hu:
> The turbulent flow is using the kernel correction only for non-surface particles. The alpha factor of linear correction is set to 1.0. Yet there are still several points that I feel confusing:
> 
> 1. `LinearGradientCorrectionWithBulkScope` is defined as this:
> 
> ```
>  Matd operator()(size_t index_j, size_t index_i)
>     {
>         return within_scope_(index_j) ? B_[index_j] : B_[index_i];
>     };
> 
>     Matd operator()(size_t index_i)
>     {
>         return B_[index_i];
>     };
> ```
> 
> which means that in the 1st half integration, if index_i and index_j are both surface particles, the correction matrix of a surface particle will still be used. Also, for contact with wall, the correction matrix of a surface particle might still be used.
> 
> 2. They use the correction matrix for boundary condition, which means that the correction matrix of buffer particles will be applied when computing the velocity change.

@FengWang3119 Could you explain this to Weiyi? Thanks.

FengWang3119:
> > The turbulent flow is using the kernel correction only for non-surface particles. The alpha factor of linear correction is set to 1.0. Yet there are still several points that I feel confusing:
> > 
> > 1. `LinearGradientCorrectionWithBulkScope` is defined as this:
> > 
> > ```
> >  Matd operator()(size_t index_j, size_t index_i)
> >     {
> >         return within_scope_(index_j) ? B_[index_j] : B_[index_i];
> >     };
> > 
> >     Matd operator()(size_t index_i)
> >     {
> >         return B_[index_i];
> >     };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which means that in the 1st half integration, if index_i and index_j are both surface particles, the correction matrix of a surface particle will still be used. Also, for contact with wall, the correction matrix of a surface particle might still be used.
> > 
> > 2. They use the correction matrix for boundary condition, which means that the correction matrix of buffer particles will be applied when computing the velocity change.
> 
> @FengWang3119 Could you explain this to Weiyi? Thanks.

Sure.
This part only considers the internal flow, no free surface

Xiangyu-Hu:
How about ht results after the set correction matrix to identity in buffer?

WeiyiVirtonomy:
I have added the reload file and reduced end time. The PR is ready for merge once the CI tests pass.

PR #905, opened 2025-08-28
Beam model

WeiyiVirtonomy:
I haven't transplanted the code to SPHinXsys yet, but you could find the documentation and some test results in the markdown file.

I got stuck with this formulation. I wanted to reproduce the test case "4.3 out-of-plane bending of the cantilever to form a helix due to a concentrated moment and an out-of-plane force" in this reference: [test](https://doi.org/10.1002/nme.6994). However, as observed in the pure bending case, the beam goes crazy after rotating for a certain degree. Unlike the 2d case where we can constrain the motions to in-plane, this is a 3d case. 

image
 
It is also weird that the 2d pure bending case can only get a solution close to the reference for certain material parameters. The rotational angle is only related to $EI$ in this case, so as long as $EI$ is constant, the beam should be bent for a cycle.

With $GA=5000$ and $EI=100$ and $\rho=1$, the result agrees well with the analytical solution when the in-plane constraint is applied.
However, when we reduce $GA$ to 500, the simulation fails immediately. I tried to reduce the time step size by using a smaller density, but it didn't help.

image

With $GA=5e5$ and $EI=100$ and density increased to $1e6$, there is only a very small rotation:

image

By reducing the physical damping, it gets a larger deformation, but still much smaller than the analytical solution:

image

The default physical damping is calculated by the empirical formula $0.1 \sqrt{\rho * E} * r$.

### DIFF
--- a/tests/3d_examples/test_3d_simo_reissner_beam/beam_wrapper.h
+++ b/tests/3d_examples/test_3d_simo_reissner_beam/beam_wrapper.h
@@ -210,6 +210,7 @@ struct bar_simulation
                 if (dt < dt_ref / 1.1)
                 {
                     output_function(output_ite++);
+                    std::cout << "Time step size too small: " << dt << std::endl;
                     throw std::runtime_error("Time step size too small, simulation stopped.");
                 }
 
--- a/tests/3d_examples/test_3d_simo_reissner_beam/simo_reissner_beam_dynamics.h
+++ b/tests/3d_examples/test_3d_simo_reissner_beam/simo_reissner_beam_dynamics.h
@@ -35,7 +35,7 @@ class BaseSimoRelaxation : public BaseBarRelaxation
     Vec3d *Tau_;        // material tensile-shear strain
     Vec3d *Kappa_;      // material bending strain
     Vec3d *dr_ds_;      // dr_ds in global coordinates
-    Mat3d *grad_r_;     // nabla_r in global coordinates
+    Mat3d *grad_d_;     // nabla_displacement in global coordinates
     Mat3d *grad_theta_; // nabla_theta in global coordinates
 
   public:
@@ -48,9 +48,7 @@ class BaseSimoRelaxation : public BaseBarRelaxation
           Kappa_(particles_->registerStateVariableData<Vecd>("Kappa")),
           dr_ds_(particles_->registerStateVariableData<Vec3d>("DrDs", [this](size_t index_i) -> Vec3d
                                                               { return transformation_matrix0_[index_i].row(xAxis); })),
-          grad_r_(particles_->registerStateVariableData<Mat3d>("NablaR", [this](size_t index_i) -> Mat3d
-                                                               {Vec3d t1 = transformation_matrix0_[index_i].row(xAxis); 
-                                                                return t1*t1.transpose(); })),
+          grad_d_(particles_->registerStateVariableData<Mat3d>("NablaDisplacement")),
           grad_theta_(particles_->registerStateVariableData<Mat3d>("NablaTheta")) {}
 };
 
@@ -62,6 +60,7 @@ class SimoReissnerStressRelaxationFirstHalf : public BaseSimoRelaxation
     Real hourglass_control_factor_;
 
     Real *mass_;
+    Vecd *pos0_; // initial position
     Real inv_W0_;
     Real dp_;
     Vec3d *pos_hourglass_;
@@ -87,6 +86,7 @@ class SimoReissnerStressRelaxationFirstHalf : public BaseSimoRelaxation
           hourglass_control_(hourglass_control),
           hourglass_control_factor_(hourglass_control_factor),
           mass_(particles_->getVariableDataByName<Real>("Mass")),
+          pos0_(particles_->registerStateVariableDataFrom<Vecd>("InitialPosition", "Position")),
           inv_W0_(1.0 / sph_body_->getSPHAdaptation().getKernel()->W0(ZeroVecd)),
           dp_(sph_body_->getSPHAdaptation().ReferenceSpacing()),
           pos_hourglass_(particles_->registerStateVariableData<Vec3d>("PositionHourglassForce")),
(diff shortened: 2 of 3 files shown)


WeiyiVirtonomy:
@Xiangyu-Hu I have pushed the code based on the shell algorithm under `tests/3d_examples/test_3d_SPHinXsys_beam`
- Numerical damping
When there is no physical damping, with a moment of 2.5 N.m, the numerical damping does help to stabilize. The left side is the one with numerical damping, while the right one is without damping. When the right one breaks, the left one is still stable.
image
image

- Instability in the pure bending
See the video. The result can be reproduced by running `pure_bending_test`
https://github.com/user-attachments/assets/e001f0e4-0b61-4371-83ab-fc8dd4798897

Xiangyu-Hu:
@WeiyiVirtonomy have you tried the same problem but with (much) high resolution?

Xiangyu-Hu:
> > @WeiyiVirtonomy have you tried the same problem but with (much) high resolution?
> 
> I increased the resolution by 4 times, but it didn't help image

Thanks for the information. This confirms that the issue is not due to resolution.

Xiangyu-Hu:
@WeiyiVirtonomy https://doi.org/10.1016/S0045-7825(98)00249-7 a paper on dynamical simulation.

Xiangyu-Hu:
@WeiyiVirtonomy Also have you tries to slowly increase the load as in the references? https://doi.org/10.1016/j.ijsolstr.2024.112746

WeiyiVirtonomy:
> @WeiyiVirtonomy Also have you tries to slowly increase the load as in the references? https://doi.org/10.1016/j.ijsolstr.2024.112746

I have tried to increase the load step by step, and only apply the next level of load when the beam has reached the steady state with the physical damping, but it didn't help

WeiyiVirtonomy:
> @WeiyiVirtonomy seems a bug in tangential operator.

Thanks, I'll try this!

Xiangyu-Hu:
> physical damping

I am confused. Physical damping should not modify the placement as it only works on velocity.

https://global-sci.com/article/91505/an-explicit-multi-time-stepping-algorithm-for-multi-time-scale-coupling-problems-in-sph

You can measure the total kinetic energy so that an increased load only applied when the equilibrium is achieved.

Xiangyu-Hu:
By the way, I think that we may use quaternion (which is actually quite easy) to update rotation.

WeiyiVirtonomy:
> By the way, I think that we may use quaternion (which is actually quite easy) to update rotation.

I will check this later.
As for the physical damping, the displacements in this example [Fig3](https://doi.org/10.1016/j.jcp.2022.111105) with a random choice of 1.0 are also much smaller than the reference, but I have this problem with a parameter of 0.5
image

WeiyiVirtonomy:
I have replaced the exponential map with the quaternion, based on [Simo](https://doi.org/10.1016/0045-7825(86)90079-4), equation (B.2)

```
      Real q0 = cos(0.5 * dtheta.norm());
        Real factor = dtheta.norm() > Eps ? sin(0.5 * dtheta.norm()) / dtheta.norm() : 0.5;
        Vec3d q_vec = factor * dtheta;
        Eigen::Quaternion<Real> q(q0, q_vec.x(), q_vec.y(), q_vec.z());
        lambda_[index_i] = q.toRotationMatrix() * lambda_[index_i];
        pseudo_b_n_[index_i] = lambda_[index_i] * b_n0_[index_i];
        pseudo_n_[index_i] = lambda_[index_i] * n0_[index_i];
```

WeiyiVirtonomy:
# Dynamic test
@Xiangyu-Hu I used a most often used dynamic example. It can be found in e.g. [Simo, VU-Quoc](10.1016/0045-7825(88)90073-4).

An out-of-plane force acts on the elbow point with the loading history shown in the figure.
image

Note that the properties are rescaled, so there doesn't exist a set of parameters that can satisfy those material properties. Here I use density=83, E=8.3e7, and r=0.063 to satisfy some of the properties.

To avoid the problems caused by a sudden change in the normal directions, I smoothed the bi-normal:
Screenshot 2025-11-03 152518

However, the elbow has very strong hourglass. It looks as if two layers of beams are teared apart:
image

To ensure this is not caused by the straight angle at the elbow, I replaced it by a straight beam, with the same loading acting on the central point, and it showed the same behavior:
image

I think it's time for us to look at the hourglass control of beam. I mimicked the shell formulation, but apparently it's not working properly.

```
    std::pair<Vec3d, Vec3d> compute_hourglass_control_force(size_t index_i)
    {
        Vec3d pos_hourglass = Vec3d::Zero();
        Vec3d theta_hourglass = Vec3d::Zero();

        const Neighborhood &inner_neighborhood = inner_configuration_[index_i];
        for (size_t n = 0; n != inner_neighborhood.current_size_; ++n)
        {
            size_t index_j = inner_neighborhood.j_[n];

            Vecd e_ij = inner_neighborhood.e_ij_[n];
            Real r_ij = inner_neighborhood.r_ij_[n];
            Real weight = inner_neighborhood.W_ij_[n] * inv_W0_;
            Vecd pos_jump = thin_structure_dynamics::getLinearVariableJump(
                e_ij, r_ij, pos_[index_i], grad_r_[index_i], pos_[index_j], grad_r_[index_j]);
            Real limiter_pos = SMIN(2.0 * pos_jump.norm() / r_ij, 1.0);
            pos_hourglass += hourglass_control_factor_ * weight * G0_ * pos_jump * Dimensions *
                             inner_neighborhood.dW_ij_[n] * Vol_[index_j] * limiter_pos;

            Vecd theta_jump = thin_structure_dynamics::getLinearVariableJump(
                e_ij, r_ij, rotation_[index_i], grad_theta_[index_i], rotation_[index_j], grad_theta_[index_j]);
            Real limiter_theta = SMIN(
                2.0 * theta_jump.norm() / ((rotation_[index_i] - rotation_[index_j]).norm() + Eps), 1.0);
            theta_hourglass += hourglass_control_factor_ * weight * G0_ * theta_jump * Dimensions *
                               inner_neighborhood.dW_ij_[n] * Vol_[index_j] * limiter_theta;
        }
        return std::make_pair(pos_hourglass, theta_hourglass);
    }
```

WeiyiVirtonomy:
@Xiangyu-Hu Sorry, there was a bug in the wrapper that caused the hourglass to switch off. Now the straight beam with a force acting on the mid-point is stable, with a period velocity on the mid-point:
image

However, the elbow one has a much larger displacement under the same loading, and finally breaks:
image
image

Xiangyu-Hu:
> @Xiangyu-Hu Sorry, there was a bug in the wrapper that caused the hourglass to switch off. Now the straight beam with a force acting on the mid-point is stable, with a period velocity on the mid-point: image
> 
> However, the elbow one has a much larger displacement under the same loading, and finally breaks: image image

I guess the elbow one introduced a singularity and formulation should able to handle it.

PR #958, opened 2025-12-18
Instability in FSI

WeiyiVirtonomy:
I added a 3D FSI case that constantly fails to discuss the instability.

The test case uses a velocity inlet and a zero-pressure outlet.
The diameter of the vessel: 25mm
The thickness of the leaflets: 1mm

image

The inlet flow is a cosine function, with a period of 1s and a maximum average inlet velocity of 0.35m/s. The inlet velocity profile is set as parabolic.
image

The Young's modulus of the leaflets: 16200 Pa
The Possion ratio: 0.49
A Saint-Venant-Kirchhoff material model is used.

- dp_f = 1mm, dp_s = 0.25mm
Under this combination of resolution, the leaflets can have a velocity up to 0.9m/s at the beginning, when the inlet maximum velocity is only 0.004m/s:
image
Despite the high initial velocity, it survived the initialization phase and opened successfully:
image
It can fail randomly, but it is more stable than the finer resolution one.

-dp_f = 0.5mm, dp_s=0.25mm
This refined fluid always fails at the beginning:

image
image


Xiangyu-Hu:
@WeiyiVirtonomy the issue seems not encountered by us. Could you identify where is the staring point of issue. You can use restart functions to locate the exact step.

PR #964, opened 2025-12-23
add gdb extension in tools/

Jasmine969:
I have developed a GDB extension for SPHinXsys in `tools/`. It supports a structured and consice format of SPHinXsys objects, e.g., `ParticleVariables` objects. In the original version, the `ParticleVariables` object is like the following in GDB (VSCode):

image

which takes users at least three clicks to get the variable names. Also, it is hard for users to get the data types and all the variable names.

With this extension, it has a consice format with unnecessary information removed (users care less about `data_field`, `device_only_variable_`, ...). At the first click, we can easily find the data types of all vectors:

image

At the second click, all the variables of a single vector can be easily found:

image

This is expected to be more convenient for users to debug, and know what variables they can get access to and output. To add this extension in your GDB, follow the instructions in tools/gdb/printers.py. It has been tested on CentOS 7.9.

Though it only supports `ParticleVariables` objects for the time being, new features can be easily added with the aid of LLM (I used GPT-5.2 for development). Really hope it will be useful to SPHinXsys.

### DIFF
--- a/tools/gdb/printers.py
+++ b/tools/gdb/printers.py
@@ -1,0 +1,446 @@
+"""SPHinXsys GDB pretty-printers (Eigen3-style).
+
+This file is intentionally written in the same style as Eigen's official
+printers: register a single lookup function via `obj.pretty_printers.append(...)`.
+
+Target: SPH::ParticleVariables / BaseParticles::all_discrete_variables_
+which is a std::tuple of std::vector<SPH::DiscreteVariable<...>*>.
+
+To use it:
+* Create a directory and put the file as well as an empty __init__.py in 
+  that directory.
+* Create a ~/.gdbinit file, that contains the following:
+    python
+    import sys
+    sys.path.insert(0, '/path/to/sphinxsys/source/dir')
+    from tools.gdb.printers import register_SPHinXsys_printers
+    register_SPHinXsys_printers(None)
+    end
+
+With this installed you can do in GDB:
+  (gdb) p all_discrete_variables_
+in VSCode,
+  (gdb) water_block.getBaseParticles().all_discrete_variables_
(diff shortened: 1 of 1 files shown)


Jasmine969:
> Could you add a small documentation on how it works?

**What it prints**

Primary target:
    - `BaseParticles::all_discrete_variables_` (aka "ParticleVariables")
    - conceptually: `std::tuple<std::vector<SPH::DiscreteVariable<...>*>...>`

**How it works (short)**

GDB chooses a pretty-printer by calling `lookup_function(val)`:
    - It checks the *type string* (after `strip_typedefs()`) and does a best-effort
        match for ParticleVariables (tuple-like structures that contain
        `SPH::DiscreteVariable`).
    - When it detects ParticleVariables, it returns `SPHParticleVariablesPrinter`,
        which yields children labeled by `_TYPE_LABELS` (index 0..N).
    - For vectors of DiscreteVariable pointers it returns
        `SPHDiscreteVariableVectorPrinter`, which iterates elements and labels them
        as `i (name, data_size=...)` when possible.
    - For `SPH::DiscreteVariable<...>*` it returns
        `SPHDiscreteVariablePtrPrinter`, which tries to show `Entity::name_` (via the
        public `Name()` accessor when available) and `data_size_`.

Iteration strategy:
    - Prefer `gdb.default_visualizer(val)` (libstdc++ STL printer) to iterate
        tuples/vectors.
    - If STL printers are not available, fall back to libstdc++ internal layout
        for `std::vector` and a minimal recursive tuple descent.
    - A global guard (`_DISABLE_OUR_LOOKUP`) prevents infinite recursion when we
        call `gdb.default_visualizer()` from inside our own printer.

PR #973, opened 2026-01-09
Two sided fluid-shell interaction

WeiyiVirtonomy:
Reset shell particle mass


WeiyiVirtonomy:
## Example: 2d aortic valve

Thickness 0.16mm, [10.1016/S0021-9290(00)00068-3](https://doi.org/10.1016/S0021-9290(00)00068-3)

-0.01s

image

image


-0.19s

image

image

-0.35s

image

image


-0.53s

image

image


-0.84s

image

image

The comparison is not so clear, as the reference didn't provide the exact velocity value, and the length of the valve in their plots seems to be shorter than the valve listed in the table. 

The fluid has velocity singularities near the shell; I wonder what the cause is.

WeiyiVirtonomy:
Fixed the density summation.

The new results:

## Example: 2d aortic valve

Thickness 0.16mm, [10.1016/S0021-9290(00)00068-3](https://doi.org/10.1016/S0021-9290(00)00068-3)

-0.01s

image

image


-0.19s

image

image

-0.35s

image

image


-0.53s

image

image


-0.84s

image

image

Xiangyu-Hu:
Could it possible to have a regression test for some observer or reduced data?

WeiyiVirtonomy:
> Could it possible to have a regression test for some observer or reduced data?

I'll see if I can find a validation case with reference data. If not, I'll use the regression test

Xiangyu-Hu:
> > Could it possible to have a regression test for some observer or reduced data?
> 
> I'll see if I can find a validation case with reference data. If not, I'll use the regression test

Thanks. Xiangyu

WeiyiVirtonomy:
I was trying to reproduce the result of this 3d validation case: [M Glück](https://www.sciencedirect.com/science/article/abs/pii/S0889974603000069)

The case has been studied in many papers, e.g. [Gao](https://www.sciencedirect.com/science/article/abs/pii/S0045782524004353?via%3Dihub)

image

To reduce computational cost, I cut the domain. The reference domain is [0,33L]x[0,8L],[0,5L], while the one I used is [3L,10L]x[2L,5.6L]x[0L,5L]

The truncated domain:

image

By using dp_f=dp_s = 0.25L/4, the result deviates significantly from the reference ones:

image

The displacement is more than two times higher than the reference. 

On a coarse resolution, the smaller domain doesn't seem to have much effect on the result. I'll increase the particle number to see if it makes a big difference.

WeiyiVirtonomy:
## Doc of the 3d validation case
- Computational time

The particle number and computational time: (**note that the simulation end time is different**)

| dp_f | dp_f/dp_s | fluid particle number | end time [sec]| computational time [sec]| 
|------|------------|------------------------|------------|-----------------------|
| 0.25L/ 4 | 1        |519680                        |  11           |  3627.8 |
| 0.25L/ 8 |  1       |4121600                      |  4             |16202.9 |

- Resolution

With an increasing particle number, the peak value is getting closer to the reference. With dp=0.25L/10, the peak value has reduced to 0.14, but I don't think we need such a high resolution to get a decent result. This paper [Gao](https://www.sciencedirect.com/science/article/abs/pii/S0045782524004353?via%3Dihub) uses the SPH method with 0.25L/6 in the refinement zone, and already agrees well with the reference.

image

WeiyiVirtonomy:
## 2D validation case

I use a simplified aortic valve as the 2d validation case: See 4.2.1 [10.1016/j.cma.2024.117634](https://doi.org/10.1016/j.cma.2024.117634)

image
image

The tip displacement of the beam agrees well with the reference at the peak, except for the oscillations at the first cycle, while the valley is a bit lower.

image
image

**Note on convergence study!!**
The problem is that the resolution has not yet converged. With an increasing number of particles, the displacements still appear to be growing and to overshoot the reference value.

Xiangyu-Hu:
> I was trying to reproduce the result of this 3d validation case: [M Glück](https://www.sciencedirect.com/science/article/abs/pii/S0889974603000069)
> 
> The case has been studied in many papers, e.g. [Gao](https://www.sciencedirect.com/science/article/abs/pii/S0045782524004353?via%3Dihub)
> 
> image
> To reduce computational cost, I cut the domain. The reference domain is [0,33L]x[0,8L],[0,5L], while the one I used is [3L,10L]x[2L,5.6L]x[0L,5L]
> 
> The truncated domain:
> 
> image
> By using dp_f=dp_s = 0.25L/4, the result deviates significantly from the reference ones:
> 
> image
> The displacement is more than two times higher than the reference.
> 
> On a coarse resolution, the smaller domain doesn't seem to have much effect on the result. I'll increase the particle number to see if it makes a big difference.

It seems the deformation is much bigger. What is the problem for it?

WeiyiVirtonomy:
I tried the single shell validation in the SPH reference. A pressure of 50Pa is applied to the shell. The displacement is also much higher than the reference, so I probably misunderstood the meaning of some parameters in their paper.

Reference, the valley is at about 0.25m:

image

My result: more than 0.3m with dp = 0.25L/7

image

I thought the problem might be related to the constraint region. Initially, I extended the plate at the bottom and fixed the particles with z < 0. If I fixed one layer particle near the bottom instead, the displacement does become smaller (though still higher than 0.25m), but the previous one should be closer to the real boundary constraint

image

image

WeiyiVirtonomy:
I haven't run the FE simulation of shell yet, but the fsi result is close to the reference _Three-dimensional simulation of a flapping flag in a uniform flow_(DOI: https://doi.org/10.1017/S0022112010000248)

The initial deviation is because I enabled the fsi only after the flow developed to avoid start-up instabilities. It will be closer to the reference data if the fsi is enabled from the beginning, but the simulation sometimes crashes.

image

Xiangyu-Hu:
> I haven't run the FE simulation of shell yet, but the fsi result is close to the reference _Three-dimensional simulation of a flapping flag in a uniform flow_(DOI: https://doi.org/10.1017/S0022112010000248)
> 
> The initial deviation is because I enabled the fsi only after the flow developed to avoid start-up instabilities. It will be closer to the reference data if the fsi is enabled from the beginning, but the simulation sometimes crashes.
> 
> image

Great. 
I think that after the flow becomes stable the result should fit anyway.

PR #1032, opened 2026-03-15
add body_part_boolean and body_part_variable

Jasmine969:
This pull request introduces a new set of classes and methods for flexible particle tagging and boolean operations on particle body parts, along with a comprehensive test case to validate these features. The changes enable tagging particles based on variable ranges and combining particle groups using union, intersection, and difference operations.

New particle tagging and boolean body part features (see `src/shared/bodies/base_body_part.*`):

* Added `BodyPartByParticleVariableRange` (with specializations for `Real` and `int`) to allow tagging of particles whose scalar variable values fall within a specified range. 
* Introduced `BodyPartByParticleBoolean` and its derived classes: `BodyPartByParticleUnion`, `BodyPartByParticleIntersection`, and `BodyPartByParticleDifference`, which enable boolean combinations (union, intersection, difference) of multiple `BodyPartByParticle` objects.

Testing and validation:

Added a new test case `test_body_part_boolean.cpp` (in `tests/2d_examples/`) that creates several particle body parts based on variable ranges and tests union, intersection, and difference operations, checking the expected particle counts and writing results to file.


Jasmine969:
The original `BodyPartByParticleVariableRange` and `BodyPartByParticleBoolean` have been removed. Now, only one body part class template is added, together with a criteria class template.

I added a new class template `BodyPartByParticleCustom`, allowing users to provide customized criteria to tag particles. A common criterion is tagging by the range of a single variable, so I implemented it in the source code, named `VariableRangeTagCriteria`, so that users do not have to define their own function object, which is more convenient. If users select particles by variable, they can directly use either `BodyPartByRealVar` or `BodyPartByIntVar`, which are aliases of `BodyPartByParticleCustom<VariableRangeTagCriteria<Real>>` and `BodyPartByParticleCustom<VariableRangeTagCriteria<int>>`

For example,

```cpp
BodyPartByRealVar part_x_plus_2y(water_block, "PartXPlus2Y", "XPlus2Y", 0.5, 1.5);
BodyPartByRealVar part_x_multiply(water_block, "PartXMultiplyY", "XMultiplyY", 0.1, 0.4);
```

If complicated logic is required, users can also define their own criterion, which should be a function object, receiving a particle index and returning whether it should be tagged.

For example,

```cpp
class UnionCriteria
{
  public:
    UnionCriteria(BaseParticles &base_particles)
        : x_plus_2y_(base_particles.getVariableDataByName<Real>("XPlus2Y")),
          x_multiply_y_(base_particles.getVariableDataByName<Real>("XMultiplyY")) {}

    bool operator()(size_t index_i) const
    {
        return (0.5 <= x_plus_2y_[index_i] && x_plus_2y_[index_i] <= 1.5) ||
               (0.1 <= x_multiply_y_[index_i] && x_multiply_y_[index_i] <= 0.4);
    }

  private:
    Real *x_plus_2y_;
    Real *x_multiply_y_;
};

BodyPartByParticleCustom<UnionCriteria> part_union(water_block, "UnionCriteria");
```

> In this way, you do not need data type template, as it can be in TagCriteria .

I still think `BodyPartByParticleCustom<VariableRangeTagCriteria<int>>` and `BodyPartByParticleCustom<VariableRangeTagCriteria<Real>>` are useful. I guess these two are enough for most users.

Jasmine969:
**Refactor: Generalize `BodyPartByParticle` with zero-overhead template criteria and safe particle binding**

# Description / Motivation

This update refactors the `BodyPartByParticle` class to natively support custom tagging criteria directly within the base class, eliminating the need for a separate `BodyPartByParticleCustom<T>` template wrapper. 

Previously, users had to manually pass `base_particles` to their criteria constructors from the outside. This resulted in redundant code and introduced a significant risk of accidentally binding the criteria to the wrong `BaseParticles` container, leading to out-of-bounds or memory errors. 

By introducing a modern C++ Duck-typing (SFINAE) architecture, the base class now automatically and securely injects its own `base_particles_` into the criteria during initialization, completely decoupling the criteria logic from the container retrieval.

# Key Changes Core Highlights

- **Automatic `BaseParticles` Injection (SFINAE)**:
  Introduced the `has_setupBaseParticles` type trait. If a user-provided criteria defines `void setupBaseParticles(BaseParticles &)`, the `BodyPartByParticle` constructor automatically calls it to wire up variable pointers securely. This guarantees that the criteria strictly operates on the `Body`'s valid particle set.
  
- **Zero-Overhead Abstraction**:
  The new `template <typename TagCriteria> BodyPartByParticle(SPHBody&, TagCriteria)` constructor leverages Template Argument Deduction (CTAD). We avoided `std::function` for the particle loop, ensuring that the compiler can easily inline the `operator()` without taking any virtual dispatch hits.

- **Unified and Simplified User API**:
  Users can now define pure-logic criteria structures and instantiate them gracefully without explicit template brackets or container passing:
  ```cpp
  // Before:
  // BodyPartByParticleCustom<UnionCriteria> part(water_block, "Name", UnionCriteria(water_block.getBaseParticles()));

  // Now:
  BodyPartByParticle part_union(water_block, UnionCriteria()); 
  ```

- **Built-in Variable Range Criteria Optimization**:
  `BodyPartByRealVar` and `BodyPartByIntVar` are now dramatically simplified, securely wrapping `VariableRangeTagCriteria` and delegating the logic to the generalized base constructor.

# Backward Compatibility
**100% Backward Compatible.** The traditional default constructor `BodyPartByParticle(SPHBody &sph_body)` is completely preserved. Legacy derived classes relying on geometric taggers (e.g., `BodySurface`, `BodyRegionByParticle`, `AlignedBoxByParticle`) require zero modifications and behave exactly as before.

# Testing
- `test_2d_body_part_variable` (test_body_part_boolean.cpp) has been updated to utilize the new template API.
- Replaced hard-coded subclass boolean operations with the generalized custom `TagCriteria` structs (`UnionCriteria`, `IntersectionCriteria`, etc.).
- Verified that all boolean subset particle counts perfectly match expected baseline counts remain precisely correct.

Xiangyu-Hu (changes requested):
Generally, the new class need to address confusion and consistency with previous code.

Jasmine969:
> Generally, the new class need to address confusion and consistency with previous code.

Thanks for the clarification. I have refactored the implementation to generalize the existing BodyPartByParticle concept through a constructor overload, instead of introducing a separate body-part type. I also cleaned up the test/documentation wording to make this explicit and avoid confusion with a new class.

Xiangyu-Hu:
> > Generally, the new class need to address confusion and consistency with previous code.
> 
> Thanks for the clarification. I have refactored the implementation to generalize the existing BodyPartByParticle concept through a constructor overload, instead of introducing a separate body-part type. I also cleaned up the test/documentation wording to make this explicit and avoid confusion with a new class.

OK. We can try to merge after the conflict is solved.

Xiangyu-Hu (approved):

PR #1082, opened 2026-06-22
Xiangyu/include oneapi exclusive scan

Xiangyu-Hu:
@yyc84 May be the one DPL can improve the performance on update configuration when the case is large.

PR #1088, opened 2026-07-12
:bug: Fix SYCL particle-sorting failures caused by excessive register…

Jasmine969:
## Description

This PR fixes a SYCL radix sort compatibility issue on NVIDIA GPUs.

The previous implementation used `sycl_queue.getWorkGroupSize()` in

https://github.com/Xiangyu-Hu/SPHinXsys/blob/0a037901d0bebb54e5fd055fb1f0b6b9fa50dd72/src/src_sycl/shared/common/algorithm_primitive_sycl.cpp#L20

However, `sycl::queue` does not provide this API in standard SYCL implementations. Using the device maximum work-group size as a replacement is also inappropriate because the maximum supported size does not consider kernel resource requirements, such as register usage.

For example, on NVIDIA RTX 3090, a work-group size of 512 for the radix sort kernel exceeds the available register budget and causes:

```cpp
sycl::_V1::nd_range_error:
Exceeded the number of registers available on the hardware.
```

## Changes

- Replace `sycl_queue.getWorkGroupSize()` with `execution_instance.getWorkGroupSize()`.
- Use the work-group size managed by the SPHinXsys execution framework.
- Fix the conditional compilation logic for `SPHINXSYS_USE_ONEDPL_SORTING`.

## Validation
Tested with:
- NVIDIA GeForce RTX 3090
- Intel oneAPI DPC++ 2024.0
- SYCL CUDA backend

The following test runs successfully:

```bash
tests/tests_sycl/3d_examples/test_3d_dambreak_sycl/bin/test_3d_dambreak_sycl --regression=yes
```

### DIFF
--- a/src/src_sycl/shared/common/algorithm_primitive_sycl.cpp
+++ b/src/src_sycl/shared/common/algorithm_primitive_sycl.cpp
@@ -2,7 +2,7 @@
 
 #include "sphinxsys_variable_sycl.hpp"
 
-#ifdef SPHINXSYS_USE_ONEDPL_SORTING
+#if SPHINXSYS_USE_ONEDPL_SORTING
 #include <oneapi/dpl/algorithm>
 #include <oneapi/dpl/execution>
 #endif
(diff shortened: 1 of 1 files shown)


Xiangyu-Hu:
@Jasmine969 Could you considering combining your change together with the pull request #1082 ? Thanks.

Jasmine969:
> @Jasmine969 Could you considering combining your change together with the pull request #1082 ? Thanks.

Done.

Xiangyu-Hu:
> > @Jasmine969 Could you considering combining your change together with the pull request #1082 ? Thanks.
> 
> Done.

Now. The github ci has issues. One is that the cache issue, i.e. cached files may but bot valid and leading memory issue. second is that the vcpkg we are used is too old so that, in windows, the vcpkg is not able to build. The first issue can be solved by delete the CI caches. But, we need some time to resolve the second issue as it may leads to other compiling issue.

PR #1090, opened 2026-07-13
Fix SYCL radix sort work-group sizing

Jasmine969:
## Description

This PR applies the SYCL radix-sort fix on top of #1082 , as suggested in #1088 .

The previous implementation uses `sycl_queue.getWorkGroupSize()`, which is not a standard `sycl::queue` API. This change instead uses the work-group size managed by the SPHinXsys execution framework:

```cpp
execution_instance.getWorkGroupSize()
```

It also changes:

```
#ifndef SPHINXSYS_USE_ONEDPL
```

to:

```
#if !SPHINXSYS_USE_ONEDPL
```

because SPHINXSYS_USE_ONEDPL is defined as a 0/1 compile-time value.

Finally, `wait()` is replaced with `wait_and_throw()` for proper propagation of asynchronous SYCL exceptions.

### DIFF
--- a/src/src_sycl/shared/common/algorithm_primitive_sycl.cpp
+++ b/src/src_sycl/shared/common/algorithm_primitive_sycl.cpp
@@ -10,9 +10,9 @@ void RadixSort::sort(const ParallelDevicePolicy &ex_policy, UnsignedInt size, Un
     UnsignedInt *index_permutation = dv_index_permutation_->DelegatedData(ex_policy);
     UnsignedInt *begin = dv_sequence_->DelegatedData(ex_policy) + start_index;
 
-#ifndef SPHINXSYS_USE_ONEDPL
+#if !SPHINXSYS_USE_ONEDPL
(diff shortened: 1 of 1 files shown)


Xiangyu-Hu:
Now. The github ci has issues. One is that the cache issue, i.e. cached files may but bot valid and leading memory issue. second is that the vcpkg we are used is too old so that, in windows, the vcpkg is not able to build. The first issue can be solved by delete the CI caches. But, we need some time to resolve the second issue as it may leads to other compiling issue.

Jasmine969:
> Now. The github ci has issues. One is that the cache issue, i.e. cached files may but bot valid and leading memory issue. second is that the vcpkg we are used is too old so that, in windows, the vcpkg is not able to build. The first issue can be solved by delete the CI caches. But, we need some time to resolve the second issue as it may leads to other compiling issue.

Thanks for the clarification. Understood. I will wait for the CI and vcpkg issues to be resolved before looking further into any remaining failures.

PR #1098, opened 2026-08-10
two tests case influenced

### LINKED ISSUE #1094: Particle instability and splashing near the outlet in test_2d_mixed_poiseuille_flow_sycl
**Description**

When running the SYCL test case test_2d_mixed_poiseuille_flow_sycl, I observed a particle instability near the outlet.

The flow remains generally stable inside the channel. However, when the fluid particles approach the outlet, the particle distribution becomes strongly distorted, and some particles are ejected from the outlet region instead of leaving the computational domain smoothly.

The same issue was also reproduced by Dr. Dong Wu on a different computer, suggesting that it is not specific to a single machine or hardware configuration.

**Expected behavior**

The fluid particles should pass through the outlet smoothly while maintaining a reasonably regular particle distribution and a stable velocity profile.

**Actual behavior**

The particle distribution becomes disordered near the outlet. A noticeable deformation of the fluid domain develops, followed by particle splashing or ejection around the outlet.

**Steps to reproduce**
Build the SYCL version of SPHinXsys.
Build and run test_2d_mixed_poiseuille_flow_sycl.
Inspect the particle distribution near the outlet in the generated results.

**Additional information**
SPHinXsys branch/commit: master 9f33297dc

**Screenshot**

Image

Image

Xiangyu-Hu:
Now, the pressure boundary can predict correct pressure when linear correction is applied.

### DIFF
--- a/tests/tests_sycl/2d_examples/test_2d_mixed_poiseuille_flow_sycl/mixed_poiseuille_flow.cpp
+++ b/tests/tests_sycl/2d_examples/test_2d_mixed_poiseuille_flow_sycl/mixed_poiseuille_flow.cpp
@@ -296,8 +296,12 @@ int main(int ac, char *av[])
     StateDynamics<execution::ParallelPolicy, NormalFromBodyShapeCK> wall_normal_direction(wall); // run on CPU
     StateDynamics<MainExecutionPolicy, fluid_dynamics::AdvectionStepSetup> water_advection_step_setup(water_body);
     StateDynamics<MainExecutionPolicy, fluid_dynamics::UpdateParticlePosition> water_update_particle_position(water_body);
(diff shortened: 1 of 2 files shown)


FengWang3119:
image
The particle splash dissappears.

While I also observe at some snapshots, near outlet, there is strange particle distribution, as below, although not very severe.
 
image
image

FengWang3119 (approved):

Xiangyu-Hu:
@FengWang3119 May be there still something wrong. As boundary condition restrict velocity direction and no transport velocity correction on the free surface particles. There is no reason to show up such disorder. Could you check if these disordered particle is surface particles?

we can keep it like this and recheck if anything serious thing happen.

PR #1112, opened 2026-09-17
Regularize CPU fluid density near moving walls

HaoboL:
## Problem

The legacy CPU flexible-fish example resets fluid density with `DensitySummationFreeStreamComplex`. Near the moving, multi-resolution fish wall, the kernel support can be incomplete before the acoustic integration starts. In a reproduced failure, the unnormalized sum reset one accepted fluid state from `1065.9363` to `829.6217 kg/m^3`; the equivalent kernel completeness was `0.8296`. This later produced a nonphysical high-speed jet.

A correction applied only to particles with a current wall neighbor was also tested and rejected. When a particle crossed the wall-contact-list boundary, adjacent particles used two different density definitions. The correction therefore has to be continuous as the wall kernel weight tends to zero.

## Change

This PR adds `ShepardDensityRegularizationWithWall` to the legacy CPU fluid dynamics and uses it in the CPU flexible-fish example.

For every real fluid particle `i`, it computes

```text
N_i = rho_i V_i W(0)
      + sum_j rho_j V_j W_ij
      + sum_a rho_i V_a^0 W_ia

D_i = V_i W(0)
      + sum_j V_j W_ij
      + sum_a V_a^0 W_ia

rho_i_new = N_i / D_i
V_i_new   = m_i / rho_i_new
```

Here `j` are fluid neighbors, `a` are wall neighbors, and `V_a^0 = m_a / rho_a^0` is the wall reference volume. The wall term is a zero-normal-gradient density extension. The same expression is evaluated for all real fluid particles, so it approaches the inner-fluid Shepard interpolant continuously when wall support vanishes. There is no contact threshold, density clamp, blending coefficient, or added dissipation.

The implementation writes the result through the existing `DensitySummation` particle variable. This preserves the data contract used by free-stream indication and shape-confinement dynamics.

The current CK/SYCL fish example already combines compression summation with an all-fluid density regularization. This PR supplies a wall-aware normalized reconstruction for the legacy CPU path; it does not claim bitwise or algorithmic equivalence with the CK implementation.

## Files

- `src/shared/particle_dynamics/fluid_dynamics/density_summation.h`: public dynamics declaration and method documentation.
- `src/shared/particle_dynamics/fluid_dynamics/density_summation.cpp`: fluid and wall Shepard numerator/denominator, followed by `V=m/rho` update.
- `tests/2d_examples/test_2d_flow_stream_around_fish/2d_flow_stream_around_fish.cpp`: replaces the legacy free-stream density sum with the new dynamics.

## Validation

Isolated branch, based directly on current `master`:

- Release build with GCC 9: **pass**.
- Official fish particle relaxation, 1000 steps: **pass**.
- Official CPU fish simulation: reached the configured `1.7 s` end time and printed total compute time (`54.03 s` on 20 logical CPUs); no solver error occurred before the post-run regression comparison.
- Existing total-mechanical-energy DTW database: **expected post-run failure**. The corrected run has a time-series length outside the existing `0.8-1.2` ratio check. The database is intentionally left unchanged in this Draft PR so maintainers can review the physics change before accepting a new reference.
- `git diff --check`: **pass**.

The isolated run changed only the density operator. Geometry, material properties, active strain, sound speed, CFL values, output interval, and end time remained at the checked-in example values. The case has no policy input, policy output, action mapping, reward, training budget, or imposed trajectory bias. Its existing active-strain field is prescribed: amplitude `0.12`, frequency `4 Hz`, wavelength `3L`, startup time `0.2 s`, and the existing upper/lower phase definition. The source keeps `U_f=1 m/s` as the viscosity and sound-speed scale, while `FreeStreamVelocity::u_ref_` remains `0`, so there is no imposed inflow.

In the full diagnostic reproduction that combined this density correction with separately reviewed time-step protections, the `1.7 s` run completed all 20 registered numerical-health checks: maximum acoustic density deviation `0.668997%`, maximum Mach `0.0681769`, and zero records above `Ma=0.1`. Those combined-run numbers support the diagnosed mechanism; they are not presented as isolated-PR regression values.

## Scope and review notes

This is a simulation-only change. It contains no reinforcement learning, sensor model, controller, reward, propulsion-efficiency claim, or biological interpretation.

The PR is Draft because the existing mechanical-energy reference necessarily rejects the changed trajectory. The main review questions are the public class/API placement and whether maintainers want to regenerate that reference in this PR or in a follow-up after accepting the density treatment.

Method references:

- Colagrossi & Landrini, 2003, *J. Comput. Phys.*, [doi:10.1016/S0021-9991(03)00324-3](https://doi.org/10.1016/S0021-9991(03)00324-3).
- Adami, Hu & Adams, 2012, *J. Comput. Phys.*, [doi:10.1016/j.jcp.2012.05.005](https://doi.org/10.1016/j.jcp.2012.05.005).


Xiangyu-Hu:
@HaoboL As the CI failed, some cases make be not compatible to the modifications. The issue need to be solved.

HaoboL (commented):
fixed

Xiangyu-Hu:
@HaoboL Could you figure out when the original density evaluation is not working. I personally does not like the density average approach as it introduced large numerical dissipation. But I can change if you have strong argument.

PR #1113, opened 2026-09-17
Align CPU acoustic time steps with moving walls

HaoboL:
## Problem

The legacy CPU and CK fluid paths currently apply the force-based time-step scale at different integration levels. CK includes

```text
sqrt(4 h |Force + ForcePrior| / m)
```

in every acoustic time-step reduction, while the legacy CPU path applies the same scale only when selecting the outer advection step. A force or acceleration change during acoustic subcycling can therefore be missed by the CPU acoustic estimate.

The wall Riemann integration also constructs its dummy pressure from the relative normal acceleration

```text
(ForcePrior_i / m_i - wall_acceleration_a) dot (-e_ia)
```

but the standard acoustic estimate does not bound this moving-wall term.

## Change

This PR is independent of the density-regularization PR and starts from current `master`.

1. `AcousticTimeStep` now matches the CK placement of the force scale:

   ```text
   acoustic_scale = max(c + |v|, sqrt(4 h |Force + ForcePrior| / m))
   dt_acoustic = acoustic_CFL h / acoustic_scale
   ```

2. `AdvectionTimeStep` returns to a velocity-only reduction, matching `AdvectionTimeStepCK`.

3. A reusable `WallAccelerationTimeStep` evaluates the positive relative normal acceleration already used by `Integration1stHalfWithWallRiemann`:

   ```text
   a_wall = max_i,a(0, (ForcePrior_i/m_i - wall_acceleration_a) dot (-e_ia))
   dt_wall = wall_CFL sqrt(h / (a_wall + TinyReal))
   ```

   Its default `wall_CFL` is `0.25`. This limits the constant-acceleration displacement over one selected step to `0.5 a dt^2 = 0.03125 h` when this criterion is active.

4. The CPU flexible-fish example uses `min(dt_acoustic, dt_wall)` for each acoustic substep.

## Files

- `src/shared/particle_dynamics/fluid_dynamics/fluid_time_step.h`: CPU acoustic force data and the reusable moving-wall time-step declaration.
- `src/shared/particle_dynamics/fluid_dynamics/fluid_time_step.cpp`: CK-aligned acoustic/advection reductions and moving-wall relative-acceleration reduction.
- `tests/2d_examples/test_2d_flow_stream_around_fish/2d_flow_stream_around_fish.cpp`: applies the wall limit inside the acoustic loop.

## Validation

Isolated branch, based directly on current `master`:

- Release build with GCC 9: **pass**.
- Official fish particle relaxation, 1000 steps: **pass**.
- Official CPU fish simulation: reached the configured `1.7 s` end time and printed total compute time (`64.27 s` on 20 logical CPUs): **pass**.
- Existing total-mechanical-energy DTW comparison against all 10 checked-in reference runs: **pass**.
- `git diff --check`: **pass**.

The isolated run changes only the time-step selection described above. Density summation, geometry, material properties, active strain, sound speed, output interval, and end time remain at the checked-in example values. The case has no policy input, policy output, action mapping, reward, training budget, or imposed trajectory bias. Its existing active-strain field is prescribed: amplitude `0.12`, frequency `4 Hz`, wavelength `3L`, startup time `0.2 s`, and the existing upper/lower phase definition. The source keeps `U_f=1 m/s` as the viscosity and sound-speed scale, while `FreeStreamVelocity::u_ref_` remains `0`, so there is no imposed inflow.

## Scope and review notes

This is a simulation-only change. It contains no reinforcement learning, sensor model, controller, reward, propulsion-efficiency claim, or biological interpretation.

The force term relocation is a direct legacy-CPU/CK alignment. The moving-wall limit is separate because the relative wall acceleration appears only in the wall-contact Riemann expression. The two are submitted together here because both protect the same acoustic subcycle and the fish example exercises both paths.

The main review questions are whether `WallAccelerationTimeStep` belongs beside the general CPU fluid time-step classes and whether maintainers prefer the fish example to construct it explicitly, as done here, or compose it through another dynamics wrapper.

Implementation references in this repository:

- `src/shared/shared_ck/particle_dynamics/fluid_dynamics/fluid_time_step_ck.hpp`
- `src/shared/particle_dynamics/fluid_dynamics/fluid_integration.hpp`


Xiangyu-Hu:
@HaoboL As the CI failed, some cases make be not compatible to the modifications. The issue need to be solved.

